PR #1, opened 2018-10-02
Improve the way how log part between timestamps is retrieved using awk

pmihalcin:
Take into consideration logs where each log line doesn't start with timestamp

https://github.com/EmbedITCZ/jbehave-support/issues/8

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
@@ -1,47 +1,53 @@
 package org.jbehavesupport.core.ssh;
 
+import static java.lang.Long.valueOf;
+import static java.lang.String.format;
+import static java.lang.String.join;
+import static java.nio.charset.Charset.defaultCharset;
+import static java.time.ZoneId.of;
+import static java.time.ZonedDateTime.now;
+import static java.time.format.DateTimeFormatter.ofPattern;
+import static net.schmizz.sshj.common.IOUtils.readFully;
 import static org.springframework.util.Assert.isTrue;
 import static org.springframework.util.Assert.notNull;
+import static org.springframework.util.StreamUtils.copyToString;
+import static org.springframework.util.StringUtils.isEmpty;
 
 import java.io.IOException;
-import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
 
 import net.schmizz.sshj.SSHClient;
-import net.schmizz.sshj.common.IOUtils;
 import net.schmizz.sshj.connection.channel.direct.Session;
 import net.schmizz.sshj.transport.verification.PromiscuousVerifier;
 import net.schmizz.sshj.userauth.keyprovider.KeyProvider;
-import org.springframework.util.StringUtils;
+import org.springframework.core.io.ClassPathResource;
 
 public class SshTemplate {
-    private String timestampFormat;
-    private SshSetting sshSetting;
-
-    private RollingLogResolver rollingLogResolver;
+    private final SshSetting sshSetting;
+    private final String timestampFormat;
+    private final String getLogBetweenTimestampsCommand;
+    private final RollingLogResolver rollingLogResolver;
     private SSHClient sshClient;
 
-    private interface Commands {
-        String TIMEZONE = "timedatectl | grep zone | grep -Po '(.*zone: )\\K(\\S+)'";
-        String EPOCH_TIME = "date +%s";
-        String FIRST_LAST_LINES_BETWEEN_TIMESTAMPS = "awk '{if ($1\" \"$2 >= \"%s\")  p=1; if ($1\" \"$2 >= \"%s\")  p=0;} p { print NR }' %s | sed -n '1p;$p'";
-        String PRINT_BETWEEN_LINES = "awk 'NR >= %d && NR <= %d' %s";
+    private abstract static class Commands {
+        static final String TIMEZONE = "timedatectl | grep zone | grep -Po '(.*zone: )\\K(\\S+)'";
+        static final String EPOCH_TIME = "date +%s";
     }
 
-    private interface Constants {
-        String CMD_DELIMITER = "; ";
+    private abstract static class Constants {
+        static final String CMD_DELIMITER = "; ";
     }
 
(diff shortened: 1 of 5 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
```diff
@@ -1,109 +1,109 @@
...
 
     private interface Constants {
         String CMD_DELIMITER = "; ";
     }
 
-    public SshTemplate(SshSetting sshSetting, String timestampFormat, RollingLogResolver rollingLogResolver) {
+    public SshTemplate(SshSetting sshSetting,
+                       String timestampFormat, String timestampPosixRegex,
+                       RollingLogResolver rollingLogResolver) {
+        isTrue(!isEmpty(sshSetting.getLogPath()), "log path must not be null or empty");
```
thradec: I think this is backward incompatible change, I'd keep the original version with 3 parameters and add the new one

thradec:
I think we should recreate the original jira issue in github issue tracker and avoid using JBUS-xy

thradec:
minor: your commit use HCG user, I'm not sure if it intentional

jbehave-support-core/src/main/resources/get-log-between-timestamps.awk
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/awk -f
```
paveljandejsek: Is this file used?

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
```diff
@@ -1,109 +1,109 @@
...
-            logContents = executeCommand(getLinesCommand).trim();
-        }
-
-        return logContents;
+    private String copySingleLogFile(final ZonedDateTime startTime,
+                                     final ZonedDateTime endTime,
+                                     final String timestampPosixRegex,
+                                     final String logPath) throws IOException {
+        DateTimeFormatter formatter = ofPattern(timestampFormat);
+        String command = copyToString(new ClassPathResource("get-log-between-timestamps-template.awk").getInputStream(),
```
paveljandejsek: Might be better to move it to class level so the file wont get loaded each time the log copying is called.

paveljandejsek:
Do we think there might be a solution to this issue by leveraging the existing timestampFormat without introducing the additional timestampPosixRegex? (not saying there is, just putting it up to try to avoid adding new config parameter if we can *somehow* avoid it :slightly_smiling_face: )

pmihalcin:
> I think we should recreate the original jira issue in github issue tracker and avoid using JBUS-xy

Done: https://github.com/EmbedITCZ/jbehave-support/issues/8

pmihalcin:
> minor: your commit use HCG user, I'm not sure if it intentional

I changed it. Thanks

pmihalcin:
> Do we think there might be a solution to this issue by leveraging the existing timestampFormat without introducing the additional timestampPosixRegex? (not saying there is, just putting it up to try to avoid adding new config parameter if we can _somehow_ avoid it 🙂 )

I was thinking about it, but I couldn't figure out how to do it. If anybody knows, let me know and I'll incorporate it.

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
pmihalcin: I addressed it. Thanks for the suggestion.

paveljandejsek:
> > Do we think there might be a solution to this issue by leveraging the existing timestampFormat without introducing the additional timestampPosixRegex? (not saying there is, just putting it up to try to avoid adding new config parameter if we can _somehow_ avoid it slightly_smiling_face )
> 
> I was thinking about it, but I couldn't figure out how to do it. If anybody knows, let me know and I'll incorporate it.

I think I found a solution without introducing the regex, give me a couple of days to verify it - I will post an update once I have something working (or find out it cant be done)

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
pmihalcin: I pushed new solution which doesn't rely on timestampRegex, hence is backward compatible.

pmihalcin:
@thradec @paveljandejsek 

I pushed alternative solution which doesn't break backward compatibility.
I also addressed all comments.
Have a look again and let me know if there is something to add or merge.

Thanks

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
```diff
@@ -51,59 +57,43 @@ public SshSetting getSshSetting() {
...
-
-            // get lines between line numbers
-            String getLinesCommand = String.format(Commands.PRINT_BETWEEN_LINES, startLine, endLine, logPath);
-            logContents = executeCommand(getLinesCommand).trim();
-        }
-
-        return logContents;
+        DateTimeFormatter formatter = ofPattern(timestampFormat);
+        String getLogBetweenTimestamps = format(getLogBetweenTimestampsCommand,
+            startTime.format(formatter), endTime.format(formatter), endTime.plusYears(1).format(formatter), logPath);
```
paveljandejsek: I would also add one minute and one second just to be safe in case that the user does not use timestamp with year and just uses minutes and seconds or just seconds like in unix timestamp.

```suggestion
            ZonedDateTime upperLimitTimestamp = endTime.plusYears(1).plusMinutes(1).plusSeconds(1);
            startTime.format(formatter), endTime.format(formatter), upperLimitTimestamp.format(formatter), logPath);
```

jbehave-support-core/src/main/resources/get-log-between-timestamps-template.awk
```diff
@@ -0,0 +1,57 @@
...
+  if (endLineNumber < 0) {
+    endLineNumber = NR;
+  }
+
+  # print log segment between start line number and end line number
+  for (i = 0; i < startLineNumber; i++) {
+    # use different pointer to inputFile than the current one
+    getline line < inputFile;
+  }
+  print line;
```
paveljandejsek: Just a question - wouldn't the following:
```
for (i = 0; i <= endLineNumber - startLineNumber; i++) {
    print line;
    getline line < inputFile;
  }
```
get rid of the `print line;` between the two for cycles? Or is it a bit more complicated and we cant get rid of `print line;` twice so easily 🙂

Issue #16, opened 2018-10-05
Web action for checkbox


Mior:
There is missing web action for checkbox input type which I find useful.

mbocek:
please provide PR

Mior:
After discussion with @thradec we decided to have only one web action "SELECT" for dropdown, checkbox and radio button inspired by webdriver design, instead of multiple web actions for each type.

PR #18, opened 2018-10-05
#16 added implementation of checkbox for SELECT web action

Mior:
please provide feedback and merge if happy


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/action/SelectWebAction.java
```diff
@@ -23,9 +23,17 @@ public void perform(WebActionContext ctx) {
 
         WebElement element = findElement(ctx);
 
-        Assert.isTrue(equalsIgnoreCase(element.getTagName(), "select"), "The SELECT action is not supported on this element type: " + element.getTagName());
-
-        new Select(element).selectByVisibleText(ctx.getData());
+        if (equalsIgnoreCase(element.getTagName(), "input") && equalsIgnoreCase(element.getAttribute("type"), "checkbox")){
+            boolean checked = ctx.getData().isEmpty() || toBoolean(ctx.getData());
```
thradec: minor: I'd use some null-safe isEmpty util method to prevent NPE, there is no guarantee that getData method will return non-null value

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/action/SelectWebAction.java
```diff
@@ -23,9 +23,17 @@ public void perform(WebActionContext ctx) {
...
+
+            if (toBoolean(element.getAttribute("checked")) ^ checked){
```
thradec: minor: could you reformat the code please `){` > `)  {` (there should be `.idea\codeStyleSettings.xml` in current master branch)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/action/SelectWebAction.java
```diff
@@ -23,9 +23,17 @@ public void perform(WebActionContext ctx) {
...
+                element.sendKeys(Keys.SPACE);
+            }
+        } else if (equalsIgnoreCase(element.getTagName(), "select")){
+            new Select(element).selectByVisibleText(ctx.getData());
+        } else {
+            throw new IllegalArgumentException("The SELECT action is not supported on this element type: " + element.getTagName());
+        }
```
thradec: minor: I'd try to decrease the complexity of this method by extracting into smaller and more descriptive methods ... what about something like this?

```
if( isCheckbox(element) ) {
    handleCheckbox(element, ctx);
} else if( isSelect(element) ) {
    handleSelect(element, ctx);
} else {
```

PR #21, opened 2018-10-08
#19 Setup code coverage

mbocek:


.travis.yml
```diff
@@ -10,22 +10,28 @@ cache:
...
+      chrome: stable
     before_script:
     - "export DISPLAY=:99.0"
     - "sh -e /etc/init.d/xvfb start"
     - sleep 3 # give xvfb some time to start
-    script: mvn verify -DskipUnitTests=true -Djbehave.report.level=STORY -Pintegration-test -B -V
+    script:
+    - mvn org.jacoco:jacoco-maven-plugin:prepare-agent-integration verify org.jacoco:jacoco-maven-plugin:report-integration -DskipUnitTests=true -Djbehave.report.level=STORY -Pintegration-test -B -V
+    - curl --silent -o jbehave-support-core/target/codacy-reporter-latest.jar -L https://github.com/codacy/codacy-coverage-reporter/releases/download/4.0.3/codacy-coverage-reporter-4.0.3-assembly.jar
+    - java -jar jbehave-support-core/target/codacy-reporter-latest.jar report -l Java -r jbehave-support-core/target/site/jacoco-it/jacoco.xml
```
thradec: according to codacy [documentation](https://github.com/codacy/codacy-coverage-reporter#setup)

> In order to send multiple reports for the same language, you need to upload each report separately with the flag --partial and then notify Codacy, after all reports were sent, with the final command.

shouldn't the report from system tests be sent with flag `partial`? and then invoked the `final` command?

.travis.yml
```diff
@@ -10,22 +10,28 @@ cache:
...
 
 jobs:
   include:
   - stage: build
     script: mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
   - stage: test
-    script: mvn test -B -V
-  - stage: test
+    script:
+    - mvn org.jacoco:jacoco-maven-plugin:prepare-agent test org.jacoco:jacoco-maven-plugin:report -Dmaven.test.failure.ignore=false -B -V
```
thradec: what is the purpose of `-Dmaven.test.failure.ignore=false`? isn't it the default value?

.travis.yml
mbocek: you are right, fixed

.travis.yml
mbocek: removed

PR #26, opened 2018-10-19
Upgrade to Spring Boot 2.0.6.RELEASE

pmihalcin:
spring-boot-starter-remote-shell was removed, so I pulled necessary bits to SshConfig

It also fixes security alert whereby Boot pulls jackson dependencies with version 2.9.7

### DIFF
--- a/jbehave-support-core-test/jbehave-support-core-test-app/pom.xml
+++ b/jbehave-support-core-test/jbehave-support-core-test-app/pom.xml
@@ -51,8 +51,46 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-remote-shell</artifactId>
+            <groupId>org.crashub</groupId>
+            <artifactId>crash.cli</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.crashub</groupId>
+            <artifactId>crash.shell</artifactId>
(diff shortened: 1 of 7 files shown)


mbocek:
Very quick question: java classes regarding ssh are simply copied from spring 1.5? I don't want to review spring code :)

paveljandejsek:
Out of curiosity - did you try to look at https://github.com/anand1st/sshd-shell-spring-boot?
(If not then I might try to investigate it in the future - just dont want to waste time if we already know it wont work for us 🙂 )

pmihalcin:
> Very quick question: java classes regarding ssh are simply copied from springboot 1.5? I don't want to review spring code :)

@mbocek 
Yes, I just copied them to our codebase. It is minimalistic version of what they had there.

pmihalcin:
> Out of curiosity - did you try to look at https://github.com/anand1st/sshd-shell-spring-boot?
> (If not then I might try to investigate it in the future - just dont want to waste time if we already know it wont work for us 🙂 )

I haven't. I didn't know about it :)
This starter could be an improvement in the future. For now, my solution will suffice as it is only test app.

paveljandejsek:
> @paveljandejsek
> I haven't. I didn't know about it :)
> This starter could be an improvement in the future. For now, my solution will suffice as it is only test app.

Definitely agree, thanks for the info - if I get bored I will try to look into it some more 🙂

PR #27, opened 2018-10-22
NPE fix for step "these rows are present in the query result"

jankralik:


thradec:
due to yesterday's github outage, the travis didn't got a notification, could you force push the commit (but with some change, so the commit will be different)

PR #37, opened 2018-11-01
Throw unchecked IO exception from SshTemplate. Solution of breaking c…

AmyKosinova:
…hange for migration.

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
@@ -14,6 +14,7 @@ import static org.springframework.util.StreamUtils.copyToString;
 import static org.springframework.util.StringUtils.isEmpty;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshTemplate.java
```diff
@@ -40,14 +41,18 @@
...
-    public SshTemplate(SshSetting sshSetting, String timestampFormat, RollingLogResolver rollingLogResolver) throws IOException {
+    public SshTemplate(SshSetting sshSetting, String timestampFormat, RollingLogResolver rollingLogResolver) {
         isTrue(!isEmpty(sshSetting.getLogPath()), "log path must not be null or empty");
         isTrue(!isEmpty(timestampFormat), "timestamp format must not be null or empty");
 
         this.sshSetting = sshSetting;
         this.timestampFormat = timestampFormat;
         this.rollingLogResolver = rollingLogResolver;
-        this.getLogBetweenTimestampsCommand = copyToString(new ClassPathResource("get-log-between-timestamps-template.awk").getInputStream(), defaultCharset());
+        try {
```
paveljandejsek: We could probably get around this with `@SneakyThrows(IOException.class)` - but might get a worse resulting exception type than `UncheckedIOException` as a result - IMO worth the payoff in saving try/catch block, what do others think?

AmyKosinova: I vote for specific wrapper before generic.

paveljandejsek: I am merging this as it is since it seems no one seems to care strongly about this. Thanks for the patience and the PR 🙂

PR #55, opened 2018-12-30
build(pom): updated libraries versions

### LINKED ISSUE #54: Libraries upgrade
Upgrade libraries to last version.

mbocek:
fix #54


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/webdriver/FirefoxWebDriverFactory.java
```diff
@@ -50,7 +51,7 @@ private void createFirefox47Driver() {
             if (browserDriverLocation != null) {
                 System.setProperty("webdriver.gecko.driver", browserDriverLocation);
             } else {
-                FirefoxDriverManager.getInstance().version("v0.8.0").setup();
+                FirefoxDriverManager.getInstance(DriverManagerType.FIREFOX).version("v0.8.0").setup();
```
paveljandejsek: Same question as with chrome above:
```suggestion
                WebDriverManager.firefoxdriver().version("v0.8.0").setup();
```

mbocek: good catch

PR #79, opened 2019-02-16
Updated dependencies

paveljandejsek:


paveljandejsek:
> Why this minor update without obvious need for it? No code change, no bug report, no nothing...

Why not?  Cant we improve the project without being bug report driven? 
Also did we not just agree on last triage that we want to have up to date dependencies? I explicitly asked for stance on this and AFAIK everyone agreed that we should do updates. I also mentioned that I will be doing PR like this and did not hear any objections from anyone. (IMHO we could even use something like Dependabot if we feel like it)
So what seems to be the problem? :frowning:

PR #87, opened 2019-03-20
Removed unnecessary (and possibly problematic) js call from WebElemen…

paveljandejsek:
Removed unnecessary (and problematic) js script call from WebElementLocator:
a) not all browsers have to support javascript (many mobile dont)
b) this is redundant and already covered by `driver.manage().timeouts().pageLoadTimeout(timeout, TimeUnit.SECONDS);` in our webdrivers
c) we introduced some mandatory magic constant 10s that does not respect timeout settings


paveljandejsek:
> From what I remember, it was added in MR from some team who adopt JBUS in an earlier version. So potentially it could break their tests. 

That is possible, because I think that in the past we did not have the `driver.manage().timeouts().pageLoadTimeout` in our drivers which should solve this issue. So I am willing to take the risk :slightly_smiling_face:

PR #104, opened 2019-04-16
Fixed potential issue with sql result validation

### LINKED ISSUE #84: SqlSteps result comparison should not delete values from saved result
Currently the internal method compareExampleTableVersusListOfMaps in SqlSteps makes data manipulation on the original sql result object:
`for (Map<String, Object> row : getSqlResult()) {
            row.entrySet().removeIf(r -> !upperCaseHeaders.contains(r.getKey()));`

This then causes problems if for example user wants to first validate one column in "these rows match"and afterwards save different column using "columns from query are saved"  - the column is already removed and not found.

We should either change this logic or work on a copy the entryset

paveljandejsek:
Closes #84 
Fixed potential issue with sql result validation changing the original result data.


paveljandejsek:
> 7 LOC : 1 comment 🙄

My reason was that without the comment I would not understand at first glance why the copying is happening and would have tendency to remove it (which I guess is also the reason why the copying wasnt already present in the first place).
Would you understand why its there even without the comment? If I get at "yes, its obvious, youre just a moron" from at least two people then I will happily remove it 😄

AmyKosinova:
I just didn't get the comment, until I understood the point of code...

Issue #106, opened 2019-04-19
Project is not build-able on openjdk 11


mbocek:

mbocek:
```
[ERROR] Failed to execute goal org.jvnet.jaxb2.maven2:maven-jaxb2-plugin:0.13.1:generate (client) on project jbehave-support-core-test-oxm: Execution client of goal org.jvnet.jaxb2.maven2:maven-jaxb2-plugin:0.13.1:generate failed: A required class was missing while executing org.jvnet.jaxb2.maven2:maven-jaxb2-plugin:0.13.1:generate: com/sun/xml/bind/api/ErrorListener
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>org.jvnet.jaxb2.maven2:maven-jaxb2-plugin:0.13.1
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[ERROR] urls[0] = file:/home/mbocek/.m2/repository/org/jvnet/jaxb2/maven2/maven-jaxb2-plugin/0.13.1/maven-jaxb2-plugin-0.13.1.jar
[ERROR] urls[1] = file:/home/mbocek/.m2/repository/org/jvnet/jaxb2/maven2/maven-jaxb2-plugin-core/0.13.1/maven-jaxb2-plugin-core-0.13.1.jar
[ERROR] urls[2] = file:/home/mbocek/.m2/repository/org/apache/commons/commons-lang3/3.2.1/commons-lang3-3.2.1.jar
[ERROR] urls[3] = file:/home/mbocek/.m2/repository/com/sun/org/apache/xml/internal/resolver/20050927/resolver-20050927.jar
[ERROR] urls[4] = file:/home/mbocek/.m2/repository/org/sonatype/plexus/plexus-build-api/0.0.7/plexus-build-api-0.0.7.jar
[ERROR] urls[5] = file:/home/mbocek/.m2/repository/junit/junit/4.8.1/junit-4.8.1.jar
[ERROR] urls[6] = file:/home/mbocek/.m2/repository/org/codehaus/plexus/plexus-utils/1.5.15/plexus-utils-1.5.15.jar
[ERROR] urls[7] = file:/home/mbocek/.m2/repository/org/jvnet/jaxb2/maven2/maven-jaxb22-plugin/0.13.1/maven-jaxb22-plugin-0.13.1.jar
[...]

PR #107, opened 2019-04-19
Java 11 compatibility fixes

paveljandejsek:
Solves #106 

Added openjdk11 to travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,50 @@
 language: java
-jdk:
-- oraclejdk8
 
 install: true
-dist: trusty
 sudo: required
 
 cache:
   directories:
-  - '$HOME/.m2/repository'
+  - '$HOME/.m2'
 
-stages:
-- build
-- test
-- system-test
-
-jobs:
+matrix:
   include:
-  - stage: build
-    script: mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-  - stage: test
-    script:
-    - mvn org.jacoco:jacoco-maven-plugin:prepare-agent test org.jacoco:jacoco-maven-plugin:report -B -V
-    - curl --silent -o jbehave-support-core/target/codacy-reporter-latest.jar -L https://github.com/codacy/codacy-coverage-reporter/releases/download/4.0.3/codacy-coverage-reporter-4.0.3-assembly.jar
-    - java -jar jbehave-support-core/target/codacy-reporter-latest.jar report -l Java -r jbehave-support-core/target/site/jacoco/jacoco.xml --partial
-  - stage: system-test
-    addons:
-      chrome: stable
-    before_script:
-    - "export DISPLAY=:99.0"
-    - "sh -e /etc/init.d/xvfb start"
-    - sleep 3 # give xvfb some time to start
-    script:
-    - mvn org.jacoco:jacoco-maven-plugin:prepare-agent-integration verify org.jacoco:jacoco-maven-plugin:report-integration -DskipUnitTests=true -Djbehave.report.level=STORY -Pintegration-test -B -V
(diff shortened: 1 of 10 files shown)


paveljandejsek:
Does anyone know why travis does the builds 207.2 and 207.3?:
https://travis-ci.org/EmbedITCZ/jbehave-support/builds/522705401

I would expect two builds for each stage with different JDK, but this build output for our config confuses me 😮

mbocek:
Looks like the matrix definition is the right solution. Check https://github.com/lihaoyi/mill/blob/master/.travis.yml

Hopefully, it will work.

paveljandejsek:
> Looks like the matrix definition is the right solution. Check https://github.com/lihaoyi/mill/blob/master/.travis.yml
> 
> Hopefully, it will work.

It should work, but any idea how to avoid copy pasting our whole mvn+other commands for every matrix record? (i.e. clean install two times, test + codacy two times, verify + codacy two times) 
Apart from putting it into separate .sh file that would get added into git (which is what the example does but at least I personally do not want to go that way unless there is no better option)

mbocek:
> > Looks like the matrix definition is the right solution. Check https://github.com/lihaoyi/mill/blob/master/.travis.yml
> > Hopefully, it will work.
> 
> It should work, but any idea how to avoid copy pasting our whole mvn+other commands for every matrix record? (i.e. clean install two times, test + codacy two times, verify + codacy two times)
> Apart from putting it into separate .sh file that would get added into git (which is what the example does but at least I personally do not want to go that way unless there is no better option)

No Idea... based on docu it should work by default. Just an idea: try to explicitly add jdk under each stage section.

paveljandejsek:
> > > Looks like the matrix definition is the right solution. Check https://github.com/lihaoyi/mill/blob/master/.travis.yml
> > > Hopefully, it will work.
> > 
> > 
> > It should work, but any idea how to avoid copy pasting our whole mvn+other commands for every matrix record? (i.e. clean install two times, test + codacy two times, verify + codacy two times)
> > Apart from putting it into separate .sh file that would get added into git (which is what the example does but at least I personally do not want to go that way unless there is no better option)
> 
> No Idea... based on docu it should work by default. Just an idea: try to explicitly add jdk under each stage section.

Doesn't seem to do a thing 😕 After lots of trial and error I am slowly starting to think that we will either need to repeat the commands twice or to extract them into external .sh files... (unless someone comes up with a better solution that I am missing 🙂)

paveljandejsek:
Please check and comment, once travis fixes [openjdk11 issues](https://github.com/TweetWallFX/TweetwallFX/issues/765), this PR should be good to go.

I copy pasted the stages twice in travis - not particularly happy with this, but apart of extracting the commands to external scripts that would become part of git repo this was the only solution I got working to build all stages on two JDKs

paveljandejsek:
Btw. As far as Travis goes - do we have a reason to stay on Trusty (end of standard support this month AFAIK) or should I try to migrate to Xenial?

paveljandejsek:
Ok taking previously said back - right now SNAFU, it seems like we will have to use two different distros for each java and I will also have to take a closer look at the commands as well 🤕

paveljandejsek:
Ok, should finally be working :slightly_smiling_face: 
Some notes:
- Due to Travis bugs/features we are building 8/11 on different Ubuntu versions
- Build commands were moved to external scripts in ci folder - due to problems with passing of strings and some special shell characters such as `&()`
- For java 8 we are using oracle jdk, for 11 openjdk

Please review (and when merging please squash commits into one)

mbocek:
Please add to release profile enforce to JDK8.

paveljandejsek:
Rebased, squashed, added jdk8 enforce - happy merging 🙂

Issue #82, opened 2019-02-27
Equals and NotEquals verifiers dont work well with arrays


paveljandejsek:
Currently these verifiers do not work correctly if passed for example two arrays of primitive types (e.g. byte[]), we should consider adding some special case for Arrays and comparison using something like Arrays.equals if  somehow possible

PR #111, opened 2019-04-22
Added array support for Equals and NotEquals verifiers

paveljandejsek:
Solves #82


paveljandejsek:
> Map<Class<?>, BiFunction<Object, Object, Boolean>>

That could work, but personally I find the ifs in this particular case more readable. What do others think?

PR #123, opened 2019-05-16
Added expression support for web wait condition.

### LINKED ISSUE #122: WebSteps - wait until condition should parse expression commands

paveljandejsek:
Closes #122


jbehave-support-core/src/main/java/org/jbehavesupport/core/web/WebSteps.java
```diff
@@ -172,12 +172,12 @@ public void verifyProperties(String page, ExamplesTable table) {
      *                  </ul>
      */
     @Then("on [$page] page wait until [$element] $condition")
-    public void waitUntilCondition(String page, String element, String condition) {
+    public void waitUntilCondition(String page, String element, ExpressionEvaluatingParameter<String> condition) {
```
thradec: FTR it is an incompatible change in api ... but I don't think anybody is using it

PR #126, opened 2019-06-04
[FIX #101] Instantiating inlaid JAXB elements in request factory

### LINKED ISSUE #101: Request factory does not support JAXBElements of native classes
While resolving JAXB elements which are last steps, we are unable to instantiate native generics for them.

paveljandejsek:
Fixes #101
Original commit + rename commit from me


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/126/index.xml)

PR #128, opened 2019-06-20
Changed Jaxb implementation to moxy

paveljandejsek:
Moxy supports converters for java8 data types which were working before explicitly adding glassfish jaxb runtime.
I have not added test for it since it would include changing xjb, creating new LocalDate converter, changing test story, changing test ws and potentially changing test xsd (but if someone feels strongly about it I can do it, it just did not seem worth the hassle to me).


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/128/index.xml)

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/128/index.xml)

PR #129, opened 2019-06-26
Simplified ci to use only xenial

paveljandejsek:
Simplified ci to use only xenial (and changed to openjdk8 instead of oraclejdk8)


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/129/index.xml)

PR #130, opened 2019-06-26
Duplicate ui mapping now throws IllegalStateException

xsmrcek:
Today we faced issue where tester made duplicate mapping for page and system taken one of the mappings without throwing any error.
I tested behaviour manually by adding file with duplicate mapping.


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/130/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/YamlElementLocatorParser.java
```diff
@@ -53,7 +53,8 @@ public void process(Resource elementLocatorsResource) {
                     "Unable to parse given entry [" + key + ": " + value + "] "
                         + "in file [" + elementLocatorsResource.getFilename() + "], "
                         + "the expected format is: pageName.elementName[.locatorType], "
-                        + "where locator type is optional, supported values are "
+                        + "where element name must be unique for page and "
+                        + "locator type is optional, supported values are "
                         + "css (default) and xpath", e);
```
paveljandejsek: Not related to this PR, but we should probably change the wording of this last lline since custom By values are now supported apart of the default css and xpath :slightly_smiling_face:

PR #132, opened 2019-06-28
Minor code cleanup

paveljandejsek:


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/132/index.xml)

xsmrcek:
Hey @paveljandejsek I just wonder why was file get-log-between-timestamps.awk removed. We are not using it anyme?

paveljandejsek:
> Hey @paveljandejsek I just wonder why was file get-log-between-timestamps.awk removed. We are not using it anyme?

There were two almost identical files:
1. `get-log-between-timestamps.awk` for straight up copy paste to shell not used anywhere in the code just for debug purposes (the one being deleted)
2. `get-log-between-timestamps-template.awk` - actually used in the code that still remains

Only difference between them is shebang at the begining and `%s` signs at the end of one that is then replaced with actual awk input parameters. So nothing that important that a person wanting to try it directly in shell would not be able to write in a couple of seconds/minutes

xsmrcek:
Thank you, I didn't saw second file.

PR #134, opened 2019-07-01
RegexFindVerifier and Comparing of columns

xsmrcek:
REGEX_FIND verifier is best used for in log checks. Like 
| myAppRequest.*?{CP:INDENTIFIER}.*?testedValue>{CP:TESTED_VALUE}<.*?myAppRequest | REGEX_FIND|.

Steps for columns comparing is good for comparing SQL resutls with many columns, because writing 100 columns on one line is really ugly., so it;s better to save all values from both selects to context and compare them in another step. Of course it have mutch more examples of usage.

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
@@ -8,7 +8,6 @@ public final class ExampleTableConstraints {
     public static final String NAME = "name";
     public static final String DATA = "data";
     public static final String ALIAS = "contextAlias";
-    public static final String ACTUAL_VALUE = "actualValue";
     public static final String EXPECTED_VALUE = "expectedValue";
     public static final String VERIFIER = "verifier";
     /**
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
@@ -8,6 +8,7 @@ public final class ExampleTableConstraints {
     public static final String NAME = "name";
     public static final String DATA = "data";
     public static final String ALIAS = "contextAlias";
+    public static final String ACTUAL_VALUE = "actualValue";
     public static final String EXPECTED_VALUE = "expectedValue";
     public static final String VERIFIER = "verifier";
     /**
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
@@ -14,6 +14,7 @@ public class RegexFindVerifier extends AbstractVerifier {
 
     public void verify(Object actual, Object expected) {
         Assert.notNull(actual, "Actual value must be provided");
+        Assert.notNull(expected, "Actual value must be provided");
         Matcher matcher = Pattern.compile(expected.toString()).matcher(actual.toString());
         if (!matcher.find()) {
             this.throwAssertionError("regex '%s' wasn't found in: \n%s", expected, actual);
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
@@ -14,7 +14,7 @@ public class RegexFindVerifier extends AbstractVerifier {
 
     public void verify(Object actual, Object expected) {
         Assert.notNull(actual, "Actual value must be provided");
-        Assert.notNull(expected, "Actual value must be provided");
+        Assert.notNull(expected, "Expected value must be provided");
         Matcher matcher = Pattern.compile(expected.toString()).matcher(actual.toString());
         if (!matcher.find()) {
             this.throwAssertionError("regex '%s' wasn't found in: \n%s", expected, actual);
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
@@ -1,0 +1,22 @@
+package org.jbehavesupport.core.internal.verification;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.springframework.stereotype.Component;
+import org.springframework.util.Assert;
+import static org.jbehavesupport.core.internal.verification.VerifierNames.REGEX_FIND;
+
+@Component
+public class RegexFindVerifier extends AbstractVerifier {
+    public String name() {
+        return REGEX_FIND;
+    }
+
+    public void verify(Object actual, Object expected) {
+        Assert.notNull(actual, "Actual value must be provided");
+        Matcher matcher = Pattern.compile(expected.toString()).matcher(actual.toString());
+        if (!matcher.find()) {
+            this.throwAssertionError("regex '%s' wasn't found in: \n%s", expected, actual);
+        }
+    }
+}
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
@@ -1,20 +1,17 @@
 package org.jbehavesupport.core.verification;
 
-import java.util.List;
 import java.util.Map;
 import lombok.RequiredArgsConstructor;
 import org.assertj.core.api.SoftAssertions;
 import org.codehaus.plexus.util.StringUtils;
 import org.jbehave.core.annotations.Then;
 import org.jbehave.core.model.ExamplesTable;
-import org.jbehavesupport.core.internal.parameterconverters.ExamplesEvaluationTableConverter;
 import org.jbehavesupport.core.internal.verification.EqualsVerifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import static org.jbehavesupport.core.internal.ExampleTableConstraints.DATA;
 import static org.jbehavesupport.core.internal.ExampleTableConstraints.EXPECTED_VALUE;
 import static org.jbehavesupport.core.internal.ExampleTableConstraints.VERIFIER;
-import static org.jbehavesupport.core.internal.ExamplesTableUtil.convertTable;
 
 @Component
 @RequiredArgsConstructor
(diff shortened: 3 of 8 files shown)


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/134/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private void verifyRow(Map<String, String> tableRow, Verifier verifier, String searchColumn1, String searchColumn2, SoftAssertions softly) {
+        if (softly.errorsCollected().size() >= 10) {
+            softly.assertAll();
+        }
+        softly.assertThatCode(() -> resolveVerifier(tableRow.get(VERIFIER), verifier).verify(tableRow.get(searchColumn1), tableRow.get(searchColumn2)))
+            .doesNotThrowAnyException();
+    }
+
+    private Verifier resolveVerifier(String verifierName, Verifier defaultVerifier) {
```
paveljandejsek: Just a question - dont we already do something like this at several places? Dont we have some util/component for it? (and if not it might be worth adding it)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
```diff
@@ -0,0 +1,22 @@
+package org.jbehavesupport.core.internal.verification;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.springframework.stereotype.Component;
+import org.springframework.util.Assert;
+import static org.jbehavesupport.core.internal.verification.VerifierNames.REGEX_FIND;
+
+@Component
+public class RegexFindVerifier extends AbstractVerifier {
```
paveljandejsek: Can you please add similar test as for other verifiers?

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/134/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
```diff
@@ -8,6 +8,7 @@
     public static final String NAME = "name";
     public static final String DATA = "data";
     public static final String ALIAS = "contextAlias";
+    public static final String ACTUAL_VALUE = "actualValue";
```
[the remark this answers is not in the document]
xsmrcek: I have two reasons for adding new column. 
First: data doesn't makes a sence there, as actualValue better describe what you want to do. If I change it to data or something, I would prefer to changing name of second column to somethink like data2.
Second: I'm just to lazy to refactor possibly a hundred of steps in our project (as I taken step from there). But If I remove Given annotation as you suggested in another comment, I will have to refactor it anyway (or keep old method anotated with Given in project to avoid refactor at all).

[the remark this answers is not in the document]
paveljandejsek: Second: ctrl shift r should do the trick if needed :slightly_smiling_face:

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,48 @@
...
+@Component
+public class VerificationSteps {
+
+    @Autowired
+    private VerifierResolver verifierResolver;
+
+    @Autowired
+    private ExamplesEvaluationTableConverter tableConverter;
+
+    @Given("following columns are compared: $stringTable")
```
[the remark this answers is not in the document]
AmyKosinova: It is asserting step, should be `then`

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExampleTableConstraints.java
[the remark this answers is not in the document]
AmyKosinova: It is just about mindset. If I say: "I expect my data to be verified for some value." Then it is perfectly fine reflected via `| data | expectedValue | verifier |`
If you are too lazy don't push yourself, you can always keep this functionality just in your project. On this level we have to meet framework requirements, that has to fit as much users as possible.
I vote against new column.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
```diff
@@ -0,0 +1,22 @@
...
+    public String name() {
+        return REGEX_FIND;
+    }
+
+    public void verify(Object actual, Object expected) {
+        Assert.notNull(actual, "Actual value must be provided");
+        Matcher matcher = Pattern.compile(expected.toString()).matcher(actual.toString());
```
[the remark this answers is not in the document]
xsmrcek: assertAdded

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
xsmrcek: I used test for RegexVerifier (REGEX_MATCH) and added teo more cases to better fit this verifier.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Component
+public class RegexFindVerifier extends AbstractVerifier {
+    public String name() {
+        return REGEX_FIND;
+    }
+
+    public void verify(Object actual, Object expected) {
+        Assert.notNull(actual, "Actual value must be provided");
+        Assert.notNull(expected, "Actual value must be provided");
```
paveljandejsek:
```suggestion
        Assert.notNull(expected, "Expected value must be provided");
```

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
xsmrcek: We may extend verifierResolver to offer this functionality. But it's not up to me.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/RegexFindVerifier.java
```diff
@@ -0,0 +1,22 @@
...
+    public String name() {
+        return REGEX_FIND;
+    }
+
+    public void verify(Object actual, Object expected) {
+        Assert.notNull(actual, "Actual value must be provided");
+        Matcher matcher = Pattern.compile(expected.toString()).matcher(actual.toString());
+        if (!matcher.find()) {
+            this.throwAssertionError("regex '%s' wasn't found in: \n%s", expected, actual);
+        }
```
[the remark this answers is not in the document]
paveljandejsek: IMHO not really since dot doesnt match EOL by default so you would have to turn on multiline mode or something like that as well (still doable but getting more and more ugly)

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Component;
+import static org.jbehavesupport.core.internal.ExampleTableConstraints.ACTUAL_VALUE;
+import static org.jbehavesupport.core.internal.ExampleTableConstraints.EXPECTED_VALUE;
+import static org.jbehavesupport.core.internal.ExampleTableConstraints.VERIFIER;
+import static org.jbehavesupport.core.internal.ExamplesTableUtil.convertTable;
+
+@Component
+public class VerificationSteps {
+
+    @Autowired
```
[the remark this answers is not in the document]
xsmrcek: I hope using it is as easy as it seems :D

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/134/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,46 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class VerificationSteps {
+
+    private final VerifierResolver verifierResolver;
+    private final ExamplesEvaluationTableConverter tableConverter;
+
+    @Given("following columns are compared: $stringTable")
+    @Then("following columns are compared: $stringTable")
```
paveljandejsek: `following data are compared`?

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/134/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Then("following data are compared: $stringTable")
+    public void compareRows(String stringTable) {
+        List<Map<String, String>> convertedTable = convertTable((ExamplesTable) tableConverter.convertValue(stringTable, null));
+        SoftAssertions softly = new SoftAssertions();
+        convertedTable.forEach(e -> verifyRow(e, verifierResolver.getVerifierByName(VerifierNames.EQ), DATA, EXPECTED_VALUE, softly));
+        softly.assertAll();
+    }
+
+    private void verifyRow(Map<String, String> tableRow, Verifier verifier, String searchColumn1, String searchColumn2, SoftAssertions softly) {
```
AmyKosinova: 1. As verifier is always the same (EQ) I suggest to remove parameter from signature and simple autowire specific verifier bean that will be used here. Or even better - only in `resolveVerifier` method.
2. `serachColumns` are also always the same value, why putting it into signature?
3. sorry for not pointing it out earlier, I overlooked this.

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -0,0 +1,48 @@
...
+public class VerificationSteps {
+
+    @Value("${verifier.max.assert.count:10}")
+    private int maxSoftAssertCount;
+
+    private final VerifierResolver verifierResolver;
+    private final ExamplesEvaluationTableConverter tableConverter;
+
+    @Then("following data are compared: $stringTable")
+    public void compareRows(String stringTable) {
```
AmyKosinova: Why do you use string, isn't `ExamplesTable` better so you don't need to do conversion explicitly, but leave if on jbehave?

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
xsmrcek: @AmyKosinova it is OK now?

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/134/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
AmyKosinova: Clean like queens crown :+1:

PR #136, opened 2019-07-03
Switched to constructor injection

paveljandejsek:
I left @Autowired on:
* optional dependencies
* groovy files
* abstract classes / classes where inheritance is expected - so requiredargs ctor can be used there without any hassle


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/136/index.xml)

PR #139, opened 2019-07-12
Removed unnecessary semicolons from groovy tests.

paveljandejsek:


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/139/index.xml)

PR #144, opened 2019-07-17
Improved verifier resolver

### LINKED ISSUE #135: resolveDefaultVerifier
Extend verifierResolver to be able to resolve default verifier.

Hebcak:
Verifier resolver can return default verifier, if given verifier is null.
Fixed #135

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
@@ -1,13 +1,14 @@
 package org.jbehavesupport.core.internal.verification;
 
-import java.util.List;
-
 import lombok.RequiredArgsConstructor;
 import org.jbehavesupport.core.verification.Verifier;
 import org.jbehavesupport.core.verification.VerifierResolver;
-
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.NoUniqueBeanDefinitionException;
+
+import java.util.List;
+
+import static org.springframework.util.StringUtils.hasText;
 
 @RequiredArgsConstructor
 public class VerifierResolverImpl implements VerifierResolver {
@@ -24,6 +25,10 @@ public class VerifierResolverImpl implements VerifierResolver {
                 return a;
             })
             .orElseThrow(() -> new NoSuchBeanDefinitionException(name));
+    }
+
+    public Verifier getVerifierByName(String name, Verifier verifierIfNameEmpty) {
+        return hasText(name) ? getVerifierByName(name) : verifierIfNameEmpty;
     }
 
 }
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
@@ -1,36 +1,12 @@
 package org.jbehavesupport.core.rest;
 
-import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
-import static org.jbehavesupport.core.internal.ExampleTableConstraints.ALIAS;
-import static org.jbehavesupport.core.internal.ExampleTableConstraints.DATA;
-import static org.jbehavesupport.core.internal.ExampleTableConstraints.EXPECTED_VALUE;
-import static org.jbehavesupport.core.internal.ExampleTableConstraints.NAME;
-import static org.jbehavesupport.core.internal.ExamplesTableUtil.getValue;
-import static org.springframework.util.Assert.state;
-
-import java.io.IOException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Consumer;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-
-import javax.annotation.PostConstruct;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.jayway.jsonpath.DocumentContext;
 import com.jayway.jsonpath.JsonPath;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Triple;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.jbehave.core.model.ExamplesTable;
 import org.jbehave.core.steps.Row;
 import org.jbehavesupport.core.TestContext;
@@ -38,7 +14,7 @@ import org.jbehavesupport.core.internal.ExampleTableConstraints;
 import org.jbehavesupport.core.internal.ExamplesTableUtil;
 import org.jbehavesupport.core.internal.MetadataUtil;
 import org.jbehavesupport.core.internal.SkipSslVerificationHttpRequestFactory;
-import org.jbehavesupport.core.internal.verification.VerifierNames;
+import org.jbehavesupport.core.internal.verification.EqualsVerifier;
 import org.jbehavesupport.core.report.extension.RestXmlReporterExtension;
 import org.jbehavesupport.core.verification.Verifier;
 import org.jbehavesupport.core.verification.VerifierResolver;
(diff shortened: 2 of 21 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
```diff
@@ -26,4 +27,12 @@ public Verifier getVerifierByName(String name) {
             .orElseThrow(() -> new NoSuchBeanDefinitionException(name));
     }
 
+    public Verifier getVerifierByName(String name, String defaultVerifierName) {
```
AmyKosinova: default verifier should be of appropriate type - not string.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
```diff
@@ -26,4 +27,12 @@ public Verifier getVerifierByName(String name) {
...
+        if (!isEmpty(name)) {
```
AmyKosinova: `hasText()` is more robust.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
```diff
@@ -26,4 +27,12 @@ public Verifier getVerifierByName(String name) {
...
+            return getVerifierByName(name);
```
AmyKosinova: Method throws an exception if none or more verifiers is found, shouldn't it fallback to default in that case?

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshSteps.java
```diff
@@ -136,8 +135,8 @@ private void checkDataPresence(String systemQualifier, ZonedDateTime startTime,
         return sshTemplates;
     }
 
-    private Verifier resolveVerifier(String verifierName, Verifier defaultVerifier) {
-        return (StringUtils.isNotEmpty(verifierName)) ? verifierResolver.getVerifierByName(verifierName) : defaultVerifier;
+    private Verifier resolveVerifier(String verifierName, String defaultVerifierName) {
```
AmyKosinova: Why do we need to keep this method and not call resolver directly?

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
```diff
@@ -39,6 +38,6 @@ private void verifyRow(Map<String, String> tableRow, SoftAssertions softly) {
     }
 
     private Verifier resolveVerifier(String verifierName) {
-        return (StringUtils.isNotEmpty(verifierName)) ? verifierResolver.getVerifierByName(verifierName) : equalsVerifier;
+        return verifierResolver.getVerifierByName(verifierName, VerifierNames.EQ);
```
AmyKosinova: Another unnecessary call for resolving.

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerifierNames.java
```diff
@@ -1,4 +1,4 @@
-package org.jbehavesupport.core.internal.verification;
```
AmyKosinova: Why pulling it out of internal?

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: Its verifier name thats why String.
And it is not Verifier because all of our verifiers (equals, contains etc) are in internal package so the users could not use them in their steps. From my POV its better to move VerifierNames to public API, than to expose all of our internal verifiers.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: Kinda makes sense, but also would theoretically break backwards compatibility (tests that were failing before could potentially start working for bad configuration or if user would input nonsensical verifier name, which could be misleading if it works)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: Considering that comment from Amy, how about something like:

```
return !hasText(name) ? getVerifierByName(name) : getVerifierByName(defaultVerifierName);
```

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
AmyKosinova: Spring doesn't care what name has your package as long as it is present in component scan. Default verifier must not be resolved, but provided.

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerificationSteps.java
Hebcak: We swapped the lines and the methods for greater readability

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: But you would be importing from internal, and there is no guarantee that the classes in internal wont be moved to different package, or renamed/replaced. 

E.g. we guarantee to users that some verifier with name EQ is present, but not that class org.jbehavesupport.core.internal.verification.EqualsVerifier will exist.

jbehave-support-core/src/main/java/org/jbehavesupport/core/verification/VerifierResolver.java
```diff
@@ -16,4 +15,13 @@
...
     Verifier getVerifierByName(String name);
+
+    /**
+     * Finds single registered bean by verifier name, if empty, returns a default one. Exactly one must exist otherwise {@link NoUniqueBeanDefinitionException} or {@link NoSuchBeanDefinitionException}
+     *
+     * @param name Verifier name regards bean setup, or from {@link VerifierNames}
+     * @param defaultVerifierName Name of a default verifier
+     * @return Single found verifier or the specified defauld verifier
+     */
+    Verifier getVerifierByName(String name, String defaultVerifierName);
```
thradec: personally, I don't like the contract of this new method, I can't remember that I'd see something similar anywhere else ... unfortunately, I don't have a better proposal ... I tend to the following code, which we have already there 
`verifierResolver.getVerifierByName(defaultIfEmpty(verifierName, EQ));` ... in any case, if others are happy with it, I'll shut up 😄

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/144/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshSteps.java
```diff
@@ -77,20 +76,20 @@ public void markStartTime(String startTimeAlias) {
...
     @Then("the following data are present in [$systemQualifier] log:$presentData")
     public void logContainsData(String systemQualifier, String stringTable) {
-        checkDataPresence(systemQualifier, scenarioStart, stringTable, verifierResolver.getVerifierByName(VerifierNames.CONTAINS));
+        checkDataPresence(systemQualifier, scenarioStart, stringTable, VerifierNames.CONTAINS);
     }
 
     @Then("the following data are present in [$systemQualifier] log since [$startTimeAlias]:$presentData")
     public void logContainsData(String systemQualifier, String startTimeAlias, String stringTable) {
-        checkDataPresence(systemQualifier, testContext.get(startTimeAlias), stringTable, verifierResolver.getVerifierByName(VerifierNames.CONTAINS));
+        checkDataPresence(systemQualifier, testContext.get(startTimeAlias), stringTable,VerifierNames.CONTAINS);
```
paveljandejsek: Nitpicking: formatting

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: Discussed in call - behavior is fine with just checking the name and not catching exceptions.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/VerifierResolverImpl.java
paveljandejsek: Discussed - in the end decided we will use appropriate type.

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
```diff
@@ -425,7 +425,7 @@ private void verifyResponseHeaders(HttpHeaders actualHeaders, ExamplesTable data
                 String assertionErrorMessage = "Headers don't contain " + headerKey + "\n" + actualResponseMessage;
                 assertThat(actualHeaders.containsKey(headerKey)).as(assertionErrorMessage).isTrue();
 
-                final Verifier verifier = verifierResolver.getVerifierByName(StringUtils.defaultIfEmpty(triple.getRight(), VerifierNames.EQ));
+                final Verifier verifier = verifierResolver.getVerifierByName(triple.getRight(), verifierResolver.getVerifierByName(VerifierNames.EQ));
```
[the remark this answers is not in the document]
paveljandejsek: +1 on injecting the internal implementations, should simplify the code. Good point 👍

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/144/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
```diff
@@ -100,6 +99,9 @@
     @Autowired
     private VerifierResolver verifierResolver;
 
+    @Autowired
+    private EqualsVerifier equalsVerifier;
```
AmyKosinova: Constructor injection

jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
```diff
@@ -71,6 +70,9 @@
     @Autowired
     private VerifierResolver verifierResolver;
 
+    @Autowired
+    private EqualsVerifier equalsVerifier;
```
AmyKosinova: Constructor injection

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
paveljandejsek: We dont want to do that - user extends the handler and could not use @RequiredArgsConstructor on his own handler if we would do that

PR #148, opened 2019-07-22
Split report.xslt to multiple files

Hebcak:
Closes #137

PR #155, opened 2019-07-31
Added support for directories in report uploading script.

paveljandejsek:
Needed for #148 to correctly zip and upload report folder (support for folders will also need to be added for JRS).


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/155/index.xml)

PR #156, opened 2019-08-01
Removed old TODO

Hebcak:
Info about timeouts is already written on line 33


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/156/index.xml)

Issue #91, opened 2019-04-03
Unification of SOAP|REST request|response handling via ExamplesTable


AmyKosinova:
While creating collection we use different notation **collection.order.attribute** VS **collection[order].attribute**. We should unify this.

```
Given [NameRequest] data for [TEST]:
| name                           | data  |
| name                           | test  |
| addressList.addressInfo.0.city | Praha |
```

```
When [POST] request to [TEST]/[user/] is sent with data:
| name                 | data           | contextAlias      |
| firstName            | Pedro          | FIRST_NAME        |
| lastName             | Salgado        | LAST_NAME         |
| addresses[0].country | Brazil         | ADDRESS_0_COUNTRY |
```

paveljandejsek:
This is related only to requests? Or also to responses?

AmyKosinova:
> This is related only to requests? Or also to responses?

Good question. Response verifying / parsing should be uniform as well.

paveljandejsek:
> Response verifying / parsing should be uniform as well.

And is it already? Or did you not check so we dont know?

AmyKosinova:
Responses are handled own way, so this needs to be addressed as well.

paveljandejsek:
Solved by #158 and #159

PR #158, opened 2019-08-05
Web service supports both period and bracket notations

Hebcak:
Nested lists can be now written both with brackets and periods.
Ex. both address[0].city and address.0.city are possible

Part of: #91

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
@@ -21,8 +21,10 @@ import org.springframework.ws.client.core.WebServiceTemplate;
 import org.springframework.ws.client.support.interceptor.ClientInterceptor;
 
 import javax.annotation.PostConstruct;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertNotNull;
 import static org.springframework.util.Assert.isTrue;
@@ -53,10 +55,12 @@ import static org.springframework.util.Assert.state;
  * }
  * </pre>
  */
+
 public abstract class WebServiceHandler {
 
     private static final String CONTEXT_SEPARATOR = ".";
     private static final String REQUEST_POSTFIX = "Request";
+    public static final String BRACKET_REGEX = "(.*)\\[(\\d+)\\](\\.)(.*)";
 
     @Autowired
     protected TestContext testContext;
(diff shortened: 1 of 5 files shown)


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/158/index.xml)

jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
```diff
@@ -53,10 +55,12 @@
  * }
  * </pre>
  */
+
 public abstract class WebServiceHandler {
 
     private static final String CONTEXT_SEPARATOR = ".";
     private static final String REQUEST_POSTFIX = "Request";
+    public static final String BRACKET_REGEX = "(.*)\\[(\\d+)\\](\\.)(.*)";
```
paveljandejsek: Nitpicking: do we need the selection groups if we dont use them? Might improve readability a bit

jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
```diff
@@ -91,14 +95,29 @@ private void initTemplateInternal() {
...
+        ExamplesTable checkedData = checkCollectionNotation(data);
+        overrideRequestData(request, checkedData);
     }
 
     public final void overrideRequestData(String request, ExamplesTable data) {
         endpointRegistry.validateRequest(request);
         TestContextUtil.putDataIntoContext(testContext, data, request);
     }
 
+    private ExamplesTable checkCollectionNotation(ExamplesTable data) {
```
paveljandejsek: convertCollectionNotation or something like that since its not just checking?

jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
Hebcak: From the code perspective they are unnecessary, however I used them to make sense of the regex, because it is a cluster of escaping special characters

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/158/index.xml)

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/158/index.xml)

AmyKosinova:
Please add groovy test with mixed notation. This should verify that elements are not duplicated

jbehave-support-core/src/main/java/org/jbehavesupport/core/ws/WebServiceHandler.java
```diff
@@ -91,14 +95,29 @@ private void initTemplateInternal() {
...
         endpointRegistry.validateRequest(request);
         TestContextUtil.putDataIntoContext(testContext, data, request);
     }
 
+    private ExamplesTable convertCollectionNotation(ExamplesTable data) {
+        List<Map<String, String>> newMapList = data.getRows().stream()
+            .map(map -> {
+                String name = map.get(ExampleTableConstraints.NAME);
+                if (name.matches(BRACKET_REGEX)) {
+                    String newName = name.replace("[", ".").replace("]", "");
```
paveljandejsek: Will it work for nested indexes?

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/158/index.xml)

jbehave-support-core/src/test/groovy/org/jbehavesupport/core/TestConfig.groovy
```diff
@@ -98,14 +99,14 @@ class TestConfig {
...
-        RollingLogResolver rollingLogResolver = new SimpleRollingLogResolver();
+        RollingLogResolver rollingLogResolver = new SimpleRollingLogResolver()
         return new SshTemplate(passwordSetting, env.getProperty("ssh.timestampFormat"), rollingLogResolver)
     }
 
     @Bean
     @Qualifier("TEST")
     WebServiceHandler requestFactoryTestHandler() {
-        return new WebServiceHandler() {
+        return new TestWebServiceHandler(env) {
```
paveljandejsek: Just a question - why was this change needed?

Hebcak: The `TestWebServiceHandler` is set up to acces the mock app without the need for further configuration

PR #164, opened 2019-08-07
No more asterisk imports

Hebcak:
IDEA's "optimise imports" function doesn't want to squash imports under asterisks anymore.


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/164/index.xml)

PR #170, opened 2019-08-19
JSON can now be sent with different data types than just string

Hebcak:


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/170/index.xml)

jbehave-support-core/docs/Rest-api.md
```diff
@@ -144,3 +144,28 @@ When [POST] request to [TEST]/[user/] is sent with data:
...
+If you leave the column blank, the data will be send as a string.
+
+Ex.:
+```
+| name      | data  | type    |
+| firstName | Mario | string  |
+| age       | 24    | int     |
+| height    | 1.56  | double  |
+| plumber   | true  | boolean |
+| powerup   |       | null    |
```
paveljandejsek: debate type null vs NULL_COMMAND in data or both

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
```diff
@@ -50,6 +51,54 @@
...
+        } else {
+            return String.class;
+        }
+    }
+
+    private static Type chooseType(String typeName) {
+        switch (typeName){
+            case "boolean":
+                return boolean.class;
+            case "int":
```
paveljandejsek: Debate java types (int/double) vs json type (number)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
```diff
@@ -50,6 +51,54 @@
...
+            } else {
+                return String.class;
+            }
+        } else {
+            return String.class;
+        }
+    }
+
+    private static Type chooseType(String typeName) {
+        switch (typeName){
```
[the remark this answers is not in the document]
Hebcak: I think for 5 values (possibly 4, if we choose to go with Json data types). It is unnecessary hassle to create a map somwhere and than call it. What do the others think?

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/170/index.xml)

jbehave-support-core/docs/Rest-api.md
thradec: do we need int and double? afaik json distinguish only numbers

thradec: I'd prefer using `{NULL}` in data column instead of `null` in type column (due to consistency )

Hebcak: There is a debate, whether we want to use Json data types (number) or Java data types (int and double) https://github.com/EmbedITCZ/jbehave-support/pull/170#discussion_r316084049

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/170/index.xml)

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/170/index.xml)

jbehave-support-core/docs/Rest-api.md
```diff
@@ -144,3 +144,29 @@ When [POST] request to [TEST]/[user/] is sent with data:
 | @header.Content-Type | application/json;charset=utf-8 |
 | @body                | Anything you want to send.     |
 ```
+
+#### Handling JSON data types
+All JSON data will be send as a string, unless specified otherwise. To specify a data type, use optional `type` column in your data table. These data types are supported:
+ + `string`
+ + `boolean`
+ + `number`
+ + `null` (_only through the `{NULL}` command_)
```
paveljandejsek: I would probably remove this  since its written above to use it in the type column (at least for me it seems confusing 🙂 ). The mention of null below plus the example should hopefully be enough.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
```diff
@@ -38,18 +45,50 @@
...
+                column2 = item.valueAs(column2Header, dataType);
+            }
             String column3 = item.values().containsKey(column3Header) ? item.valueAs(column3Header, String.class) : null;
             result.add(new ImmutableTriple<>(column1, column2, column3));
         });
         return result;
     }
 
+    private static Type getValueType(Parameters item, String data) {
+        if (item.values().containsKey(ExampleTableConstraints.TYPE)) {
```
paveljandejsek: Nitpicking: how about `return String.class` at the end of the method unconditionally to simplify the two if/else a bit to be a bit more readable? Or some simplification like that...

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/170/index.xml)

PR #166, opened 2019-08-09
Improved documentation

Hebcak:
Moved all info for developers from the main README.md to CONTRIBUTING.md.
Added simple setup guide for users to main README.md
Added more info to contributors guide

PR #173, opened 2019-08-19
Issue #172 fix

xsmrcek:
Report was broken in MS Edge and IE. Adding those for lines fixed the issue (those were removed in another pull request).

### DIFF
--- a/jbehave-support-core/src/main/resources/report.xslt
+++ b/jbehave-support-core/src/main/resources/report.xslt
@@ -13,6 +13,12 @@
 
     <xsl:output method="html" indent="yes"/>
 
+    <xsl:variable name="smallcase" select="'abcdefghijklmnopqrstuvwxyz'"/>
+    <xsl:variable name="uppercase" select="'ABCDEFGHIJKLMNOPQRSTUVWXYZ'"/>
+
+    <xsl:variable name="index-source" select="document('index.xml')"/>
+    <xsl:variable name="totalStories" select="count(//story)"/>
+
     <xsl:template name="textColor">
         <xsl:param name="outcome"/>
         <xsl:param name="iconize" select="false()"/>


xsmrcek:
Report was broken in MS Edge and IE. Adding those for lines fixed the issue (those were removed in another pull request).

paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/173/index.xml)

AmyKosinova:
Well, I strongly suspect, that this change cause this but... in PR #166 I can open and close clickable elements in report> http://95.168.205.146:8080/reports/166/Ssh.xml
While in this PR I can not... http://95.168.205.146:8080/reports/173/Ssh.xml
:confused:

xsmrcek:
> Well, I strongly suspect, that this change cause this but... in PR #166 I can open and close clickable elements in report> http://95.168.205.146:8080/reports/166/Ssh.xml
> While in this PR I can not... http://95.168.205.146:8080/reports/173/Ssh.xml
> 😕

This is really strange, because I can't open and close those elements in current master or some older pull requests.

jbehave-support-core/src/main/resources/report.xslt
```diff
@@ -13,6 +13,12 @@
 
     <xsl:output method="html" indent="yes"/>
 
+    <xsl:variable name="smallcase" select="'abcdefghijklmnopqrstuvwxyz'"/>
```
paveljandejsek: Do we need smallcase, uppercase and totalStories here? And I guess this will probably just overwrite index-source that is in envelope - I think that ideally we should have it in one place only a pass it as a parameter or something if possible.

xsmrcek:
It will be solved in another pull request together with another report issues.

PR #168, opened 2019-08-16
Calling xslt templates by name, not match

Hebcak:
If user wanted to use his own reporter, his own tags won`t collide with ours

PR #175, opened 2019-08-20
Fixed switching between reports

Hebcak:
Addresses issue discovered during the revision process for #168 (should be merged before)


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/175/index.xml)

AmyKosinova:
There is same issue here as mentioned in https://github.com/EmbedITCZ/jbehave-support/pull/173#issuecomment-523039145

Hebcak:
It seems to still be broken in Firefox. Working on a way to fix it

PR #179, opened 2019-08-23
Fixed Firefox

### LINKED ISSUE #177: Switching between reports doesn't work in Firefox
The dropdown menu showing all reports in the same folder doesn't show up when the arrow is clicked, even though it is generated in the html. This issue predates the great xslt split of PR #148.

Hebcak:
Fixed switching between reports and removed the text "!DOCTYPE html" in top left corner in Firefox

Fixes #177


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/179/index.xml)

PR #180, opened 2019-08-23
Fixed opening and collapsing of step details in reports

### LINKED ISSUE #167: Test steps opening and collapsing is broken
Step details for different scenarios have the same url, causing only the details in the first scenario to open/collapse

Hebcak:
Fixes #167


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/180/index.xml)

PR #183, opened 2019-08-29
Added simple guides for Rest and Web service

Hebcak:


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/183/index.xml)

PR #187, opened 2019-09-19
SshSteps.readLog() public access

xsmrcek:
I would like to make readLog() public. So projects that are using some custom ssh logic, can utilize sshSteps too.

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshSteps.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshSteps.java
@@ -70,7 +70,7 @@ public final class SshSteps {
         logCache.clear();
     }
 
-    private String readLog(String systemQualifier, ZonedDateTime startTime, ZonedDateTime endTime) {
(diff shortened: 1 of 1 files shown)


paveljandejsek:
Generated [reports](http://95.168.205.146:8080/reports/187/index.xml)

paveljandejsek:
Nice commit message :D

jbehave-support-core/src/main/java/org/jbehavesupport/core/ssh/SshSteps.java
```diff
@@ -70,7 +70,7 @@ public void init() {
         logCache.clear();
     }
 
-    private String readLog(String systemQualifier, ZonedDateTime startTime, ZonedDateTime endTime) {
+    public String readLog(String systemQualifier, ZonedDateTime startTime, ZonedDateTime endTime) {
```
mbocek: I'm still not sure that this API should be public. We should discuss it again and probably vote.

xsmrcek: This Is a reason for separate PR :)

PR #200, opened 2019-10-18
Updated travis ci to use Ubuntu 18.04 LTS

paveljandejsek:


paveljandejsek:
Looks like travis does not support older jdks on bionic, so closing (and github actions are starting to look even better now)

PR #238, opened 2019-12-03
'Given the following values are saved:' updated according issue 191

xsmrcek:


jbehave-support-core/docs/Test-context.md
```diff
@@ -15,18 +15,8 @@ Given [ClientRequest] data for [MYAPP]:
...
-```
-Given the following test constants with prefix CONSTANTS and values:
-| name                  | data               |
-| FIRST_NAME            | {RANDOM_STRING:10} |
-| LAST_NAME             | {RANDOM_STRING:15} |
-```
-
-There's also a variant of the above step without a prefix.
 After this step is executed the test context will contain two key-value pairs - FIRST_NAME, LAST_NAME.
+Data stored by this step will always have UserDefined() metadata.
```
paveljandejsek:
```suggestion
Data stored by this step will always have `userDefined()` metadata.
```

Issue #216, opened 2019-10-30
MultiBrowser testing


paveljandejsek:
We should try to run our browser tests in multiple browsers - ideally using something like browserstack/saucelabs

paveljandejsek:
Update: browserstack info added to md in #231, license was approved and granted.
Next steps: test setup + pipeline setup

paveljandejsek:
Browserstack key will be encrypted as part of simple pipeline that will be run on master only after PR is merged (so as to avoid leaking the key publicly)

paveljandejsek:
Initial setup done in #239 
Next steps:
1. Make Web.story/WebAction.story working again :tm: 
1. Change BrowserStackIT to load all stories from some folder instead of specifying them one by one (and move all ui stories to that folder)

paveljandejsek:
Tests now run in Chrome, Firefox and Safari in special pipeline. Not entirely 100% stable, but solved with retry in pipeline.

PR #239, opened 2019-12-05
Browserstack testing setup

paveljandejsek:
Initial setup of #216


paveljandejsek:
We have some issue for Java 11 ITs but other than that ready for review...

paveljandejsek:
Java 11 issues in this build should be fixed by #242

Issue #60, opened 2019-01-16
Revise WebTableSteps based on user feedback


paveljandejsek:
We should revise WebTableSteps based on user feedback (hopefully provided in the future by Amy) and decide if we want to add tests for them, rewrite them, change them or remove them alltogether.

paveljandejsek:
Seeing as we dont get any feedback from users on this topic I propose to simplify web table steps to the basic html implementation - support just standard table tags (table, tr, td, th, etc...)

paveljandejsek:
update 2021-09-01 - we should conserve current steps (so someone can reuse them as custom steps or something), and simplify to basic html

paveljandejsek:
note from old jira issues "it use custom webelement locating method, instead of WebElementLocator" - we should probably take a look at this as well

paveljandejsek:
Old steps conserved in new repo, new ones are simplified version of basic html tables. Done in v2.

PR #244, opened 2019-12-13
Deprecation documentation improvements.

### LINKED ISSUE #227: Deprecation approach
If we want to consider some minor/major release in the future we should start using "since" and "forRemoval" attributes. E.g. @Deprecated(since = "1.0.8", forRemoval = true).

Ideally we should add this info to all existing deprecated methods as well.

paveljandejsek:
Closes #227

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/web/WebTableSteps.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/web/WebTableSteps.java
@@ -30,7 +30,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 /**
- * @deprecated support for general usable table steps will be added in future version and it is not guarantee that it will be backward compatible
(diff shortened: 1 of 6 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/web/WebTableSteps.java
```diff
@@ -30,7 +30,7 @@
 import org.springframework.stereotype.Component;
 
 /**
- * @deprecated support for general usable table steps will be added in future version and it is not guarantee that it will be backward compatible
+ * @deprecated(since = "1.0.0", forRemoval = true) support for general usable table steps may be added in a future version and it is not guaranteed that it will be backwards compatible
  */
```
xsmrcek: Are those steps really 'forRemoval=true'?

paveljandejsek: Based on the current wording and parameters I presume we will eventually remove the steps alltogether, but definitely not before we have some suitable replacement in terms of functionality. So for removal makes sense, but not until we have some replacement which we dont have and dont know what it should be yet (should have been solved by #60 )

PR #248, opened 2019-12-14
Minor cleanup of groovy code.

paveljandejsek:


paveljandejsek:
Ignore the travis failure, its confused and tries to do a "push" build on branch that does not exist (push build should be run only for branches from EmbedITCZ repo not for forks). Correct one for fork is the pr one which passed.

PR #272, opened 2020-01-20
Sonar integration

xsmrcek:


.github/workflows/ci.yml
```diff
@@ -51,6 +51,11 @@ jobs:
           mvn org.jacoco:jacoco-maven-plugin:prepare-agent test org.jacoco:jacoco-maven-plugin:report -B -V
           curl --silent -o jbehave-support-core/target/codacy-reporter-latest.jar -L https://github.com/codacy/codacy-coverage-reporter/releases/download/4.0.3/codacy-coverage-reporter-4.0.3-assembly.jar
           java -jar jbehave-support-core/target/codacy-reporter-latest.jar report -l Java -r jbehave-support-core/target/site/jacoco/jacoco.xml --partial
+      - name: Upload math result for tests
```
paveljandejsek:
```suggestion
      - name: Upload test coverage for unit tests
```

.github/workflows/ci.yml
```diff
@@ -81,4 +86,34 @@ jobs:
...
+        with:
+          name: jacoco-it
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Run Maven
+        run: mvn package org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=xsmrcek_jbehave-support -Dsonar.organization=xsmrcek -Dsonar.host.url=https://sonarcloud.io -Dsonar.login=${{ secrets.SONAR_TOKEN }} -Dsonar.coverage.jacoco.xmlReportPaths=/home/runner/work/jbehave-support/jbehave-support/jacoco/jacoco.xml,/home/runner/work/jbehave-support/jbehave-support/jacoco-it/jacoco.xml
```
paveljandejsek: Just a question do we have to have absolute path to the jacoco files?
And if so can we use something like `pwd` or something like that for the first part of the path? (I am just afraid that the path on the github runners can change in the future)

xsmrcek: I had to. But well we may figurer out how to use shorter version :)

paveljandejsek: Ok, thanks 🙂 One other thing - I would add skipTests to this command so it does not run unit tests (or maybe `mvn compile` would be enough instead of `mvn package`?) But thats just nitpicking 🙂

PR #273, opened 2020-01-20
SqlSteps.compareExpectedRows test update + output fix

xsmrcek:


xsmrcek:
@AmyKosinova You were right :) I broken output of steps (and it wasn't tested before so I didn't notice).
The new test simulates the issue I was fixing.

Issue #291, opened 2020-02-13
Web&RestServices should validate duplicates in requests


xsmrcek:
We should start validating duplicate lines in Example tables for web and rest services.

PR #295, opened 2020-02-18
Check for duplicates in examples table added

xsmrcek:
Should solve issue #291 .

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
@@ -4,25 +4,31 @@ import java.lang.reflect.Type;
 import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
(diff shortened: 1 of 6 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/ExamplesTableUtil.java
```diff
@@ -164,6 +164,56 @@ public static void assertMandatoryColumns(ExamplesTable examplesTable, String ..
...
+            for (int row = 0; row < examplesTable.getRowCount(); row++) {
+                if (duplicates.contains(examplesTable.getRow(row).get(column))){
+                    continue;
+                }
+                StringBuilder sb = new StringBuilder();
+                checkColumnForDuplicates(row, sb, duplicates, examplesTable, column);
+                if (!sb.toString().equals("")){
+                    softly.fail(sb.toString() + "}");
+                }
+            }
```
xsmrcek: return sb, do not send it in.

xsmrcek: Send softly instead, and fail in method.

xsmrcek: recursion?!

paveljandejsek:
My 2c - previous version seemed much more readable, this while more functional and "pretty" makes less sense to me and takes me more time to understand 🤕

PR #304, opened 2020-02-26
Finding clickable element now requires its visibility as well.

paveljandejsek:


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebElementLocatorImpl.java
```diff
@@ -42,6 +43,7 @@ public WebElement findElement(String pageName, String elementName) {
     @Override
     public WebElement findClickableElement(String pageName, String elementName) {
         By locator = elementRegistry.getLocator(pageName, elementName);
+        waiting().until(visibilityOfElementLocated(locator));
```
thradec: In documentation for `elementToBeClickable` is _"An expectation for checking an element is __visible__ and enabled such that you can click it."_ It doesn't work as declared?

xsmrcek: I think that 'elementToBeClickable()' isn't waiting. It checks exactly the same as Pavel's new code, but doesn't wait.

PR #317, opened 2020-03-03
Format test context in reports

xsmrcek:
Added formatting for multi-line values in test context.

### DIFF
--- a/jbehave-support-core/src/main/resources/report-generator/story.xslt
+++ b/jbehave-support-core/src/main/resources/report-generator/story.xslt
@@ -218,7 +218,6 @@
                             </td>
                             <td>
                                 <xsl:call-template name="break">
-                                    <!--xsl:value-of select="value"/-->
                                     <xsl:with-param name="text" select="value"/>
                                 </xsl:call-template>
                             </td>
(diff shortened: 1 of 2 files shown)


jbehave-support-core/src/main/resources/report-generator/story.xslt
```diff
@@ -217,7 +217,10 @@
                                 <xsl:value-of select="key"/>
                             </td>
                             <td>
-                                <xsl:value-of select="value"/>
+                                <xsl:call-template name="break">
+                                    <!--xsl:value-of select="value"/-->
```
paveljandejsek:
```suggestion
```

jbehave-support-core/src/test/groovy/org/jbehavesupport/test/sample/SshReport.story
```diff
@@ -32,3 +32,11 @@ Then the following data are present in [TEST] log:
 | header     |
 | long       |
 | unexpected |
+
+Scenario: Save more values into context to demonstrate Test Context extension behaviou
```
paveljandejsek: Minor: do we want it in ssh report story or somewhere else?

xsmrcek: I was thinking about renaming SshReport story to Report stoy. We doesn't have any another sotry representing report.

Issue #326, opened 2020-03-10
RestSteps can not send empty list in request


xsmrcek:

xsmrcek:
`customerInfo: {firstName: "cvzxv", middleName: "", lastName: "vxczvz", suffix: "USA",…}
firstName: "cvzxv"
middleName: ""
lastName: "vxczvz"
suffix: "USA"
addresses: [{type: "BILLING", firstLine: "1 Aaa 3", secondLine: null, city: "Lewellen", state: "NE",…}]
phoneNumbers: [{type: "PHONE_MOBILE", value: "6841684641"}]
emails: [{email: "zcvza@fdsa.fa"}]
ssn: "646416846"
dateOfBirth: "1989-09-15"
languagePreference: "ENG"
documents: []
questionAnswers: [{code: "AQ013", values: [{code: "AQR061", value: "Unemployed"}]},…]`

PR #334, opened 2020-03-27
Added the possibility to send an empty json list in rest steps

romcooo:
Added the possibility to send an empty json list in rest steps and fixed 3 SonarLint warnings.
Solves https://github.com/EmbedITCZ/jbehave-support/issues/326

### DIFF
--- a/jbehave-support-core/docs/Rest-api.md
+++ b/jbehave-support-core/docs/Rest-api.md
@@ -53,6 +53,16 @@ When [POST] request to [TEST]/[user/] is sent with data:
 | addresses[2].city       | Graz 2           | ADDRESS_2_CITY      |
 ```
 
+It's possible to send an empty list by using empty brackets.
+In the following example, an empty list of documents will be sent:
+```
+When [POST] request to [TEST]/[user/] is sent with data:
+| name        | data   | contextAlias |
+| firstName   | Ricky  | FIRST_NAME   |
+| lastName    | Gruber | LAST_NAME    |
+| documents[] |        |              |
+```
+
 To send a list at the root level use the following syntax:
 
 ```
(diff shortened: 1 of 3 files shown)


romcooo:
>
>
> Nice first PR, good work!
> 
> Couple of thoughts from my side:
> 
>     * I know that it was like that in the previous versions and is the same with the other pattern - but are we able to replace this with something like `indexedKeyPattern3.pattern() + ".*"` to avoid repetition of the same pattern again? Not saying to do it, just wondering if its possible or if there is some problem I am not seeing at first glance. Also if possible I would probably rename the `indexedKeyPattern*` variables to something more readable (again not saying to do it in this PR, but might be worth creating issue/another PR for it)
> 
>     * This solution supports only the `[]` notation for the array in examples table not the `.` one, are we okay with that? (I dont have a clue how we would even write it for the dot case, just want to point it out that this is the case).
> 
>     * Can you add some example to the markdown documentation as well?

Thanks for the feedback :) In order:

1. I agree that the patterns could probably be made more clear, but I was a bit scared to make these changes in my first PR :) But I can take a better look at this on Monday.
2. Honestly, I forgot about the dot notation for lists, sorry about that. I think a notation such as "documents." could indicate an empty list, but it seems a bit less self-explanatory than "documents[]". I personally think that it would be acceptable to have only the "[]" notation support empty lists, but if you or others think that it would be better to have support in both, I can try to add it as well.
3. I documented the new functionality in the Rest-api.md file and amended my previous commit - please, take a look and let me know if it's sufficient.

paveljandejsek:
> 1. I agree that the patterns could probably be made more clear, but I was a bit scared to make these changes in my first PR :) But I can take a better look at this on Monday.

Dont do them just yet, I am just wondering if its possible for some future work, or if what I said is total BS 🙂 

> 2. Honestly, I forgot about the dot notation for lists, sorry about that. I think a notation such as "documents." could indicate an empty list, but it seems a bit less self-explanatory than "documents[]". I personally think that it would be acceptable to have only the "[]" notation support empty lists, but if you or others think that it would be better to have support in both, I can try to add it as well.

I dont know myself either, I cant think of any "nice" way to write it with dot notation, but I am personally fine without it - so I would maybe leave this point open till wednesday when hopefully more people will take a look at it.

> 3. I documented the new functionality in the Rest-api.md file and amended my previous commit - please, take a look and let me know if it's sufficient.

Awesome, approved by me 👍

PR #341, opened 2020-03-31
TableParameterAligner sonar issues solved

xsmrcek:


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/TableParameterAligner.java
```diff
@@ -7,6 +7,11 @@
 import java.util.List;
 
 public class TableParameterAligner {
+
+    private TableParameterAligner() {
```
paveljandejsek: Nitpick, but I would probably use `@UtilityClass` since we already use it in other places.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/TableParameterAligner.java
```diff
@@ -93,7 +92,7 @@ public static String alignTableInString(String inputString) {
 
             for (int i = 0; i < split.length; i++) {
                 String field = split[i].trim();
-                int maxFieldLength = maxColumnWidths[i];
+                int maxFieldLength = maxColumnWidths.length < i ? 0 : maxColumnWidths[i];
```
paveljandejsek: Just a question - can this occur? (And if so when?)

PR #363, opened 2020-04-22
StaleExceptionReferenceException problem solved

xsmrcek:


jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jbehavesupport.core.web.WebWaitConditionResolver;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.springframework.retry.annotation.Backoff;
+import org.springframework.retry.annotation.EnableRetry;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@EnableRetry
```
paveljandejsek, line 21: Is this needed? My guess would be config + test config should be enough.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String ELEMENT = "element";
+    public static final String PROPERTY = "property";
+
+    private final WebPropertyResolver propertyResolver;
+    private final WebWaitConditionResolver waitConditionResolver;
+    private final WebActionResolver actionResolver;
+
+    @Retryable(
+        value = { StaleElementReferenceException.class },
+        backoff = @Backoff(delay = 1000))
```
paveljandejsek, line 35: I think that 1000 delay is default, so you should be able to omit it 🙂

jbehave-support-core/src/test/groovy/org/jbehavesupport/core/web/StaleElementExceptionTest.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+import org.jbehave.core.model.ExamplesTable
+import org.jbehavesupport.core.TestConfig
+import org.jbehavesupport.core.TestContext
+import org.jbehavesupport.core.expression.ExpressionEvaluatingParameter
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.test.context.ContextConfiguration
+import spock.lang.Specification
+
+@ContextConfiguration(classes = TestConfig.class)
```
paveljandejsek, line 11: Shouldnt TestConfig or this test have `@EnableRetry`?

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
xsmrcek: it is enought to have it in config.

jbehave-support-core/src/test/groovy/org/jbehavesupport/core/web/StaleElementExceptionTest.groovy
xsmrcek: Default config is enought

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
xsmrcek: It is, thanks

jbehave-support-core/pom.xml
```diff
@@ -217,6 +217,14 @@
...
             <artifactId>spring-jdbc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-aspects</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.retry</groupId>
+            <artifactId>spring-retry</artifactId>
+        </dependency>
```
thradec, line 227: do we want to add another two dependencies due to this feature?

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        WebWaitCondition waitCondition = waitConditionResolver.resolveWaitCondition(waitConditionCtx);
+        waitCondition.evaluate(waitConditionCtx);
+    }
+
+    @Retryable(value = { StaleElementReferenceException.class })
+    public void performAction(WebActionContext actionContext, String actionName){
+        WebAction action = actionResolver.resolveAction(actionName);
+        action.perform(actionContext);
+    }
+}
```
thradec, line 53: I don't think that introducing "internal util class" is the right approach. What if somebody is using WebAction/WebProperty directly, wouldn't this be beneficial for him too? I'd say that if we add this feature it should work OOTB.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jbehavesupport.core.web.WebWaitConditionResolver;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@RequiredArgsConstructor
+@Component
+public class WebStepsUtils {
```
thradec, line 21: My main concern is if we just don't hide some missing "wait condition" behind couple of "retries", with wishful thinking "this time it will work". What if not? Will we increase the number of retries?

jbehave-support-core/pom.xml
xsmrcek: I am able to do this by ugly loop, I just thought this solution is pretty and it can be used again in different cases.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
xsmrcek: We can make it public.

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/WebStepsUtils.java
xsmrcek: Well you are right, we are doing this. But from my experiance: This exception appear only after loading page or doing some action and it is kinda rare,  3 retries with 1 sec delay should provide enought time for page to become "stable". I reader quite a lot about this exception, and answer was always the same, if you catch this exception, you have to retri whole operation including finding the element. So I wanted to provide solution so our core steps can be used without ugly static waits solving issues with this exception.
Anothe aproach I found is that element locator check if element is stale and if so refresh it, but it doesn't solve case when element become stale after being found and before it being used.

paveljandejsek: From what I read one of the "official" solutions is to just retry couple times.. If it fails even more times then I think its fine to let the test fail and let the user handle it in its own way. (so this proposed solution should be fine in that regard)

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/RefreshableWebElement.java
```diff
@@ -0,0 +1,347 @@
...
+import org.openqa.selenium.Point;
+import org.openqa.selenium.Rectangle;
+import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.remote.RemoteWebElement;
+
+import java.util.ArrayList;
+import java.util.List;
+
```
xsmrcek, line 17: javadoc

CONTRIBUTING.md
```diff
@@ -56,6 +56,9 @@ We test our code against Chrome, Safari and Firefox browsers. Tests are not runn
 
 To be able to run test in your local branch please fill your BrowserStack credentials into environment variables: `BROWSER-STACK_USERNAME`, `BROWSER-STACK_KEY` (overwrites properties in test.yml).
 
+## Using WebElements
+We use `WebElementLocatorImpl` which wraps WebElements into `RefreshableWebElement` to avoid throwing `StaleElementReferenceException`.
+If you are modifying web testing please use `WebElementLocatorImpl` or deal with `StaleElementReferenceException` in your code.
```
xsmrcek, line 62: remove

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/RefreshableWebElement.java
```diff
@@ -0,0 +1,347 @@
...
+public class RefreshableWebElement extends RemoteWebElement {
+
+    private static final int ITERATIONS = 4;
```
xsmrcek, line 20: impements all interfaces, instead of extending RemoteWebelement

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/web/RefreshableWebElement.java
```diff
@@ -0,0 +1,347 @@
...
+
+    @Delegate(excludes = WebElement.class)
+    private RemoteWebElement webElement;
+
+    private WebElementLocatorImpl webElementLocator;
+    private String webElementName;
+    private String webElementPage;
+
```
xsmrcek, line 28: final

PR #369, opened 2020-05-05
Improved BrowserStackIT + uncommented all chrome and firefox tests.

paveljandejsek:


paveljandejsek:
Split browserstack tests to three files that need to be run one at a time, fine tuned browserstack properties, added categorization to builds and story names. Browserstack test names have been changed from IT to T so they do not get picked up by failsafe automatically (if anyone has better name suggestion feel free to write what)

PR #386, opened 2020-06-08
WebProperties are checked with soft assertions now

xsmrcek:
- Web property now use soft assertions
- Sql strinct match step improved error message


jbehave-support-core/src/main/java/org/jbehavesupport/core/web/WebSteps.java
```diff
@@ -180,8 +182,10 @@ public void verifyProperties(String page, ExamplesTable table) {
             String verifierName = resolveVerifierName(values);
 
             Verifier verifier = verifierResolver.getVerifierByName(verifierName);
-            verifier.verify(actual, expected);
+            softly.assertThatCode(() -> verifier.verify(actual, expected))
```
paveljandejsek, line 185: Is it possible to add values of element and property columns to the error message? (my guess would be using the `as` method on the assertion result) - so that user would know not just the value comparison but also what line exactly threw it 🙂

xsmrcek: Very good point. I didn't thought about it as we are not doing it anywhere else.

xsmrcek:
How does it look like?
image

PR #392, opened 2020-06-17
Updated search.output.css selector in Google story

romcooo:
The old selector doesn't work anymore, I updated it in order for the scenario to work on the current version of the google search result page.


paveljandejsek:
Merged, thanks a lot @romcooo

PR #393, opened 2020-06-18
Fixed the README tutorial to suggest a test directory for story files…

romcooo:
… (previously, it implied that the tests should be in the /main/java/ directory, however, this was not aligned with another part of the tutorial which instructed the user to add a pom definition which marked the test directory as the actual location of the story files. If the user followed the tutorial strictly, it resulted in a seemingly successful test run, which, however, ended up not running any of the stories (because they would actually be in the /main/java/ directory). Also a minor English fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The `elementLocatorsSource` methods sets the name of a [file](#create-a-ui-mappi
 
 ### Create a ui mapping file
 
-In this yaml file, we need to setup the links to web page elements we want to interact with. I has to be placed in the resources directory, which is on one level above your main code directory:
+In this yaml file, we need to setup the links to web page elements we want to interact with. It has to be placed in the resources directory, which is on one level above your main code directory:
 - Project
     - src
         - main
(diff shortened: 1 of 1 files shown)


paveljandejsek:
Good points thanks - makes me think though: the test directory as defined by maven should be used for tests of the current app (ie. would make sense for tests of the test app - tests of users custom commands, handlers etc....), so we should probably change the guide to have it place everything outside of the test directory? Or not? Might be worth discussing this.

romcooo:
Another small thing I found related to your point is the fact that currently, if the directory `src/test/java` is specified in pom.xml under `build.testResources.testResource.directory`, then after running maven reimport + generate sources and update folders, it ends up marking that `java` directory as `Test Resources Root` instead of `Test Sources Root`. I'm not sure if this is fine since it contains runnable classes (it might be).

Generally, I agree with you that it's likely cleaner to have the files outside of the test directory, but after discussion with @xsmrcek, it seems that most (if not all) projects put all the stories inside the test directory, so I fixed the doc in line with that (at least for now).

Issue #398, opened 2020-06-25
Support for Splunk logs verification


stjarna:
Given that new apps get deployed to Kubernetes cluster and SSH custom step is not applicable in such case, JBUS could support custom steps related to log verification in Splunk.

It would be similar to what is already available in [SSH](https://github.com/EmbedITCZ/jbehave-support/blob/master/jbehave-support-core/docs/Ssh.md).

Instead of ssh hostnames, it would support Splunk hostname, custom index name/search expression.
Though not sure about APITokens or some kind of credentials that will let to call Splunk with technical account.

paveljandejsek:
Solved by #411 and #521

PR #411, opened 2020-07-20
Splunk Custom Steps

stjarna:
The changes relate to Splunk custom steps. The intention is to let dev teams to assert some of the events also in Splunk logs.

Please note that the test runs against mocked Splunk instance and the development was conducted with locally running Splunk Docker image.

There have been some talks with EIT Splunk Monitoring team to let us to use java sdk, but it may take a while until we are capable to use it in EIT domain.

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkClient.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkClient.java
@@ -1,0 +1,59 @@
+package org.jbehavesupport.core.splunk;
+
+import org.jbehavesupport.core.internal.splunk.SplunkOutputModes;
+
+import java.util.List;
+
+
+/**
+ * This an interface exposing search entry points for Splunk Java SDK.
+ * <p>
+ * Prior to your testing, verify that you have a valid Splunk account and you know hostname and port.
+ * <p>
+ * You can use username/password or auth token if you generated one. All the configuration elements should go to your application yaml file.
+ *
+ * <p>
+ * Example:
+ * splunk:
+ * host: <YOUR SPLUNK HOST>
+ * port: <YOUR SPLUNK HOST PORT>
+ * scheme: https
+ * credentials:
+ *  username: admin
+ *  password: password
+ * or
+ *  token: Bearer <YOUR SPLUNK AUTH TOKEN>
+ */
+
+
+public interface SplunkClient {
+
+    /**
+     * This method invokes Splunk search.
+     *
+     * @param query            Splunk search query
+     * @param earliestTime     earliest date of the search frame
+     * @param latestTime       latest date of the search frame
+     * @param splunkOutputMode Splunk output mode (e.g. "xml", "json", "csv")
+     * @return Splunk resultset matching the search query
+     */
+    List<SplunkSearchResultEntry> search(String query, String earliestTime, String latestTime, SplunkOutputModes splunkOutputMode);
+
+    /**
+     * This method invokes Splunk search.
+     *
+     * @param query        Splunk search query
+     * @param earliestTime earliest date of the search frame
+     * @param latestTime   latest date of the search frame
+     * @return Splunk resultset matching the search query
+     */
+    List<SplunkSearchResultEntry> search(String query, String earliestTime, String latestTime);
+
+    /**
+     * This method invokes Splunk search.
+     *
+     * @param query Splunk search query
+     * @return Splunk resultset matching the search query
+     */
+    List<SplunkSearchResultEntry> search(String query);
+}
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkConfig.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkConfig.java
@@ -1,0 +1,45 @@
+package org.jbehavesupport.core.splunk;
+
+import com.splunk.SSLSecurityProtocol;
+import com.splunk.ServiceArgs;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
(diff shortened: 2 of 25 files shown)


paveljandejsek:
Solves #398

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkConfig.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.NonNull;
+import org.springframework.stereotype.Component;
+
+import java.util.Objects;
+
+@Data
+@Builder
+@AllArgsConstructor
+@Component
+public class SplunkConfig {
```
paveljandejsek, line 17: It just occured to me - shouldnt we allow multiple splunk configs for multiple servers (distinguished by qualifier or something)? AFAIK in EIT there are several local splunk servers which are then aggregated in the central one, do we want to allow just one splunk connection or multiple? (though I guess real world usage will be just one splunk server)

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkSteps.java
```diff
@@ -0,0 +1,98 @@
...
+    private void verifySearchResultEntries(ExamplesTable verifiersData, List<SplunkSearchResultEntry> splunkSearchData) {
+        convertTable(verifiersData).forEach(row ->
+            assertThat(splunkSearchData)
+                .as("all found Splunk search result data meet all verifier definitions")
+                .allSatisfy(
+                    splunkSearchResultEntry -> {
+                        String verifierName = row.get(VERIFIER);
+                        String verifierData = row.get(DATA);
+                        Verifier verifier = verifierResolver.getVerifierByName(verifierName);
+                        if (Objects.nonNull(verifier)) {
```
paveljandejsek, line 85: should not be needed - default verifier resolver should throw exception if its not found (altough user can redefine it by its own resolver)

pom.xml
```diff
@@ -455,6 +464,11 @@
                 <artifactId>browserstack-local-java</artifactId>
                 <version>${version.browserstack-local-java}</version>
             </dependency>
+            <dependency>
+                <groupId>com.splunk</groupId>
+                <artifactId>splunk</artifactId>
+                <version>${version.splunk}</version>
+            </dependency>
```
thradec, line 471: Hi, my two cents (I talked about it with Pavel, so he should have more context)
- splunk indexing is asynchronous 
- splunk dependency should be optional
- there should be strict split between api and implementation
- we should not expect any particular fields in events (eg. level, namespace)

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkSteps.java
stjarna: Ok, I will remove as it is handled by verifierResolver.getVerifierByName() whose validations I did not check before.

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkClient.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param earliestTime earliest date of the search frame
+     * @param latestTime   latest date of the search frame
+     * @return Splunk resultset matching the search query
+     */
+    public List<SplunkSearchResultEntry> search(String query, String earliestTime, String latestTime) {
+        return search(query, earliestTime, latestTime, String.join(",", SPLUNK_SEARCH_FIELDS), JSON);
+    }
+
+    private void validateTimes(String earliestTimeStr, String latestTimeStr) {
+        ZonedDateTime earliestTime = nonNull(earliestTimeStr) ? ZonedDateTime.parse(earliestTimeStr, DATETIME_FORMATTER) : null;
```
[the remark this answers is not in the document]
stjarna: You are right, it will work fine since the method without format uses internally ISO_OFFSET_DATE_TIME + some optional part.

I will remove it.

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkClient.java
```diff
@@ -0,0 +1,157 @@
...
+ * credentials:
+ * username: admin
+ * password: password
+ * or
+ * token: Bearer <YOUR SPLUNK AUTH TOKEN>
+ */
+
+@RequiredArgsConstructor
+@Slf4j
+@Component
```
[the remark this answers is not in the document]
stjarna: Ok, no problem moving it JBehaveDefaultConfig. Interface will be provided too.

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkConfig.java
stjarna: As discussed with Pavel, we will stick currently with a single Splunk instance. Should there be a need to have it env specific, guys can use Spring profile.

pom.xml
stjarna: Thanks for your input. I talked to Pavel and the comments regarding the raised concerns are below:

- splunk indexing is asynchronous
  If there is a latency which is likely case, we will have to wait before we conduct the search.

- splunk dependency should be optional
  OK

- there should be strict split between api and implementation
  OK

- we should not expect any particular fields in events (eg. level, namespace)
  Ok, field preselection was removed. Splunk API will return all the fields that it provides.

stjarna:
FYI, I was also discussing with Pavel possibility to use @ConditionalOnProperty or @ConditionalOnExpression that would check on the presence of particular Splunk property in env in case of SplunkClient instantiation.

Both annotations are from spring.boot.autoconfigure and we decided that we will skip it as well as we wont also implement own conditional.

Issue #494, opened 2020-10-29
Check if we can move YamlPropertiesConfigurer loading to earlier phase


paveljandejsek:
We should check if we can move YamlPropertiesConfigurer loading to earlier phase or introduce some new property file that would be loaded at the earlier phase.
That way we could start to use env properties for bean configuration in conditionals, in constructors etc...

paveljandejsek:
We cannot move it without doing breaking change - since currently it is used as a bean and we need (ideally) properties from it for the conditional bean loading which has two phases:
```
ConfigurationPhase {
        PARSE_CONFIGURATION,
        REGISTER_BEAN;
```
Which beans that as a bean its too late for it to have effect.
I propose to keep the old implementation for backwards compatibility, mark it as deprecated and create a new yaml property loading system next to it (or let user deal with it alltogether with `@PropertySource` or something like that)

Issue #429, opened 2020-08-05
POC - easier user configuration/autoconfiguration 


paveljandejsek:
based on properties, beans etc....

paveljandejsek:
done in #791

PR #505, opened 2020-11-15
Rewritten yaml properties loading.

paveljandejsek:
Open points: 
- package and naming for the new class
- deprecation warning
- do we want to deprecate or do we want to keep both ways and just mention that one is not suitable for config (and just test data and similar stuff)


paveljandejsek:
Solves #494, should make #429 easier

xsmrcek (approved):

PR #511, opened 2020-11-19
Introduced new Env command.

### LINKED ISSUE #414: Command to support env variables
We should provide some command to allow user to load environment variables (e.g. ENV)
This could be useful for using in step expressions.

paveljandejsek:
Closes #414 

Open points:
- do we want the name to be ENV?
- do we want to handle case sensitivity, different cases like boot does?
- do we want to allow null return or throw exception?


paveljandejsek:
Dan:
- ENV ok
- Nice to have not neccessary to handle case sensitivity etc?
- Throw exception for null value

Amy:
- name ok
- would be good to have case sensitivity, camel case etc support
- Throw exception for null value

xsmrcek (approved):

thradec (approved):

PR #521, opened 2020-12-07
Splunk report extension

xsmrcek:
New extensions for splunk.
How does it look?
image

There is more attributes in splunk result entity (level, message ,namespace ,index, host) but I don't see a reason to put them to report extension (namespace and index is going to be part of query most probably and rest isn't interesting).

### DIFF
--- a/jbehave-support-core/src/main/resources/report-generator/splunkXmlReport.xslt
+++ b/jbehave-support-core/src/main/resources/report-generator/splunkXmlReport.xslt
@@ -1,0 +1,83 @@
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+
+    <xsl:template name="splunk">
+        <xsl:param name="storyIndex"/>
+        <xsl:for-each select="splunk">
+            <p>
+                <div class="card">
+                    <xsl:attribute name="id">
+                        <xsl:value-of select="concat('splunk-queries',$storyIndex)"/>
+                    </xsl:attribute>
+                    <div class="card-header">
+                        <i class="fa fa-file-text-o" aria-hidden="true"></i>
+                        Splunk queries (<a href="#splunk-queries-body" id="expand-all-splunk-queries">Toggle all queries</a>)
+                        <a href="#splunk-queries-body" data-toggle="collapse" class="float-right">Collapse</a>
+                    </div>
+                    <div id="splunk-queries-body" class="card-body collapse show">
+                        <xsl:choose>
+                            <xsl:when test="searchResult">
+                                <xsl:call-template name="searchResult"/>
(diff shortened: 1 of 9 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkSearchContext.java
```diff
@@ -0,0 +1,13 @@
+package org.jbehavesupport.core.splunk;
```
paveljandejsek, line 1: Minor, but I would probably put it in internal? (I know that right now rest is not there and sql is - we should probably align that in the future - one way or the other 🙂)

jbehave-support-core/src/main/resources/report-generator/splunkXmlReport.xslt
```diff
@@ -0,0 +1,83 @@
...
+                        <i class="fa fa-file-text-o" aria-hidden="true"></i>
+                        Splunk queries (<a href="#splunk-queries-body" id="expand-all-splunk-queries">Toggle all queries</a>)
+                        <a href="#splunk-queries-body" data-toggle="collapse" class="float-right">Collapse</a>
+                    </div>
+                    <div id="splunk-queries-body" class="card-body collapse show">
+                        <xsl:choose>
+                            <xsl:when test="searchResult">
+                                <xsl:call-template name="searchResult"/>
+                            </xsl:when>
+                            <xsl:otherwise>No SPLUNK queries available</xsl:otherwise>
```
paveljandejsek, line 21: Nitpick but why uppercase SPLUNK 😄

paveljandejsek (commented):
Looks fine to me, only open thing I have (and this will probably be up to some general brainstorming about how to do it since its not relevant only to report but to implementation as well): When I use splunk I am often interested in some other fields that will be different for each project (e.g. someone can have stack trace in lets say `json.stack_trace`, someone can have it in `message`, I might be interested in kubernetes pod name and stuff like that)

But this is definitely not part of this MR, just wanted to point it out 🙂

jbehave-support-core/src/main/resources/report-generator/splunkXmlReport.xslt
xsmrcek: why not :) But I fixed it

jbehave-support-core/src/main/java/org/jbehavesupport/core/splunk/SplunkSearchContext.java
xsmrcek: Ok it went to internal

paveljandejsek (approved):

mbocek (approved):

PR #542, opened 2021-01-09
Fixed saving of rest headers from response if body is empty.

### LINKED ISSUE #541: Rest step savig response fails
When there is no response body, and user tryes to save headers, stepm fails:

String response = testContext.get(REST_RESPONSE_JSON).toString();
HttpHeaders headers = testContext.get(REST_RESPONSE_HEADERS);

Cannot invoke "Object.toString()" because the return value of "org.jbehavesupport.core.TestContext.get(String)" is null

paveljandejsek:
Closes #541

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
@@ -403,11 +403,18 @@ public class RestServiceHandler {
     }
 
     public void saveResponse(ExamplesTable mapping) {
-        String response = testContext.get(REST_RESPONSE_JSON).toString();
+        boolean responsePresent = testContext.get(REST_RESPONSE_JSON) != null;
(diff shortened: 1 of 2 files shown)


xsmrcek (approved):

paveljandejsek:
Question - do we want to fail or return null when header that is not present is being saved? Current fix makes the value being saved in context null (instead of previous NPE)

thradec:
I'd say, it should behave the same way as for saving parts of the response body, when the path is missing.

paveljandejsek:
> I'd say, it should behave the same way as for saving parts of the response body, when the path is missing.

Good point, currently we do not solve it on our side and let json path implementation handle it, eg:
`com.jayway.jsonpath.PathNotFoundException: No results for path: $['asdName']`

So I added assert to the headers... Do we want to add assert to json path as well (to keep the exceptions the same?)

thradec:
> Do we want to add assert to json path as well (to keep the exceptions the same?)

I think it is not necessary.

thradec (approved):

Issue #463, opened 2020-09-24
Support JsonPath for REST steps and xpaths for WS


xsmrcek:
We should support using JSonPath and xpaths in our steps. It allows more complex testing and solve some issues users may have.

PR #557, opened 2021-01-19
Added JSONPath verification support

paveljandejsek:
One of possible solutions for #463 (JSONPath part)

Open points:

- Is this limited support (user must always start with `$.` if they want to use JSONPath) enough? Or do we want to do it in some other more complicated way (`@json.path` prefix, new step, or something completely different)?
- Does anyone know why IDEA always escapes the dollar sign and if thats a bug or not? (does not seem to be doing this on GitHub though)

image


paveljandejsek:
Might be worth to add some step to determine size (both X and JSON path)

xsmrcek (approved):

mbocek (approved):

Issue #661, opened 2021-07-14
Number formatting issue


xsmrcek:
It may happend that we save number in scientific format to context. We need a way out of it. <br>
I would propose command to reformat string representation of number  ('1.0E7' to '10000000') <br> 
and NUMERIC_EQUALS verifier which would convert anything to bigDecimals and then compare it (string like 1.0E7, 10000000, 10000000.00 are not equal but numbers are).

Demonstare issue with saving to context:
```json
[
    {
        "code": "29821207",
        "type": "Two weelers",
        "category": "Two weelers",
        "producerCode": "HONDA",
        "producerText": null,
        "serialNumber": "391947",
        "modelNumber": null,
        "engineNumber": null,
        "licencePlateNumber": null,
        "price": 25000000,
        "imei": null
    },
    {
        "code": "29821208",
        "type": "Other Two Wheelers",
        "category": "Two weelers",
        "producerCode": "YAMAHA",
        "producerText": null,
        "serialNumber": "14006991",
        "modelNumber": null,
        "engineNumber": null,
        "licencePlateNumber": null,
        "price": 10000000,
        "imei": null
    }
]
```

>Then response values from [TEST] REST API are saved: 
>| name           | contextAlias |
>| $.[0].price    | TEST_VALUE   |

produce: '25000000'

```json
{
    "contracts": [
        {
            "contractNumber": "4100903790",
            "type": "CEL",
            "contractStatusCode": "A",
            "loanAmount": {
                "currency": "VND",
                "value": 17587635
          },
            "instalmentAmount": {
                "currency": "VND",
                "value": 810000
            },
            "signatureDate": "2021-05-18",
            "lastInstalmentDate": "2024-05-18",
            "term": 36,
            "interestRate": 0.354862,
            "productCode": "TW17AD",
            "productBusinessDescription": "TW ULTIMATE DL",
            "processingFee": null,
            "safePayFee": null,
            "initTransactionType": "POS",
          "commodities": [
                {
                    "code": "29821207",
                    "type": "Two weelers",
                    "category": "Two weelers",
                    "producerCode": "HONDA",
                    "producerText": null,
                    "serialNumber": "391947",
                    "modelNumber": null,
                    "engineNumber": null,
                    "licencePlateNumber": null,
                    "price": 25000000,
                    "imei": null
                },
                {
                    "code": "29821208",
                    "type": "Other Two Wheelers",
                    "category": "Two weelers",
[...]

PR #687, opened 2021-11-10
Mathematical eqality verifier (issue #661)

romcooo:
Added a new MATH_EQ verifier which attempts to convert the values to a double, then compares them.
Related to issue #661

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/MathEqualsVerifier.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/MathEqualsVerifier.java
@@ -1,0 +1,35 @@
+package org.jbehavesupport.core.internal.verification;
+
+import org.springframework.stereotype.Component;
+
+import static org.jbehavesupport.core.verification.VerifierNames.MATH_EQ;
+
+@Component
+public final class MathEqualsVerifier extends AbstractVerifier {
+
+    public static final String MESSAGE_PARSE_OK = "value %s (originally written as '%s') is not mathematically equal to %s (originally written as '%s')";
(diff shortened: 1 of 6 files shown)


xsmrcek (approved):

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/MathEqualsVerifier.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    public void verify(Object actual, Object expected) {
+        if (actual == expected) {
+            return;
+        }
+
+        try {
+            Double actualDouble = Double.valueOf(actual.toString());
```
paveljandejsek, line 25: More of a question, but cant the toString hurt us in some way? Maybe conversionService.convert might be a bit safer in this case?

paveljandejsek (approved):
Thanks a lot for this!!! Happy to see you here 🙂 

Open question - now that we have MATH_EQ, is there something else we would want MATH_ version of? @xsmrcek as a requestor what do you think?

jbehave-support-core/src/main/java/org/jbehavesupport/core/internal/verification/MathEqualsVerifier.java
romcooo: @paveljandejsek  You're probably right about the conversionService. I'll update it accordingly.

Regarding the other MATH_ versions, the only one I can think of for now is MATH_NEQ (eg. `1.0 NEQ 1 == true`, but `1.0 MATH_NEQ 1 == false`) - @xsmrcek what do you think?

PR #768, opened 2023-03-15
Migrate to JUnit5

paveljandejsek:
Migrate to JUnit5 - both in unit tests using jupiter engine and in story files using the custom jbehave engine.

Removed vintage engine (junit4), backwards incompatible change in health check steps since it leaked junit4 API, replaced with soft assertion.

**Snyk is angry due to selenium downgrade - will be fixed later by #769**

### DIFF
--- a/jbehave-support-core/src/main/java/org/jbehavesupport/core/healthcheck/HealthCheckSteps.java
+++ b/jbehave-support-core/src/main/java/org/jbehavesupport/core/healthcheck/HealthCheckSteps.java
@@ -1,15 +1,12 @@
 package org.jbehavesupport.core.healthcheck;
 
-import static java.util.stream.Collectors.toList;
-
-import java.util.ArrayList;
 import java.util.List;
 
 import lombok.RequiredArgsConstructor;
+import org.assertj.core.api.SoftAssertions;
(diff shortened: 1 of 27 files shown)


jbehave-support-core/src/main/java/org/jbehavesupport/core/healthcheck/HealthCheckSteps.java
```diff
@@ -48,26 +45,23 @@ public final class HealthCheckSteps {
...
                 healthCheck.check();
             } catch (Exception e) {
-                IllegalStateException ie = new IllegalStateException("Component " + componentQualifier + " is not healthy. " + e.getMessage(), e);
-                exceptions.add(ie);
+                softly.fail("Component " + componentQualifier + " is not healthy. " + e.getMessage(), e);
             }
         }
 
-        if (!exceptions.isEmpty()) {
-            throw new MultipleFailureException(exceptions);
```
paveljandejsek, line 69: We should probably mention this incompatible API change somewhere, but dont have any idea where... Apart of release notes when releasing new version I cant think of anything else - any other ideas or are release notes enough?

xsmrcek (approved):

PR #774, opened 2023-04-26
SpringBootTest introduced

michalfranek:
Making AbstractSpringStories SpringBootTest.
Custom ConditionalOnMissingBean removed in sake of Spring Boot implementation


paveljandejsek (approved):
approved, but would merge once pipeline passes (once we have newer junit runner that fixes the reflection errors)

PR #791, opened 2023-12-26
Introduce reporting autoconfiguration.

paveljandejsek:
TODO/Open points:

- update migration.md (mention that we no longer need reporting beans explicitly declared)
- update reporting.md (mention newly introduced properties)
- cleanup commented out testconfig
- test if it really works like expected 🙂 
- agree if the new property names are ok like this?
- do we want configurationproperties for the new properites or just loose string based coupling?
- do we want to expose the new autoconfig condition (the prefix one)


jbehave-support-core/src/main/java/org/jbehavesupport/core/XmlReporterAutoConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+import java.lang.annotation.Target;
+import java.util.Arrays;
+import java.util.Map;
+
+@Configuration
+public class XmlReporterAutoConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnProperty(name = "reporting.enabled", havingValue = "on", matchIfMissing = true)
```
xsmrcek, line 47: Shouldn't property `enabled` have values `true/false`?

jbehave-support-core/src/main/java/org/jbehavesupport/core/XmlReporterAutoConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+    @ConditionalOnMissingBean
+    @ConditionalOnPropertyPrefix(prefix = "environmentInfo")
+    public EnvironmentInfoXmlReporterExtension environmentInfoXmlReporterExtension(Environment environment) {
+        return new EnvironmentInfoXmlReporterExtension(environment);
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnBean({RestServiceHandler.class, XmlReporterFactory.class})
+    public RestXmlReporterExtension restXmlReporterExtension(TestContext testContext, FileNameResolver fileNameResolver) {
```
xsmrcek, line 87: I would add conditional on property with default true, to give users option to turn it off.

xsmrcek (changes requested):

jbehave-support-core/src/main/java/org/jbehavesupport/core/XmlReporterAutoConfiguration.java
paveljandejsek: Done, I have added several (hopefully havent forgotten any)

xsmrcek (approved):

JakubVychodil (approved):

PR #881, opened 2026-04-22
Add support for Java 25 and Spring Boot 4

Mejdlo:
Add support for java 25 and spring boot 4


JakubVychodil (approved):

pom.xml
```diff
@@ -69,14 +69,16 @@
     </developers>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>21</java.version>
 
         <!-- enable annotation processing for lombok in java 23: https://github.com/projectlombok/lombok/issues/3752 -->
         <maven.compiler.proc>full</maven.compiler.proc>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
```
paveljandejsek: Do we really need to set source and target explicitly? Isn't the `<maven.compiler.release>${java.version}</maven.compiler.release>` set by Boot parent enough for our cases?

jbehave-support-core/src/test/resources/db/init.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS PERSON (
```
paveljandejsek: Shouldn't this be handled by the already existing [changelog](https://github.com/EmbedITCZ/jbehave-support/blob/master/jbehave-support-core-test/jbehave-support-core-test-app/src/main/resources/db/changelog/db.changelog-master.yaml) instead?

jbehave-support-core/src/test/groovy/org/jbehavesupport/test/TestConfiguration.java
```diff
@@ -69,12 +71,21 @@ public WebServiceHandler testWebServiceHandler() {
     @Bean
     @Qualifier("TEST")
     public DataSource testDatasource() {
-        DataSourceBuilder dataSourceBuilder = DataSourceBuilder.create();
-        dataSourceBuilder.driverClassName(env.getProperty("db.driver"));
-        dataSourceBuilder.url(env.getProperty("db.url"));
-        dataSourceBuilder.username(env.getProperty("db.username"));
-        dataSourceBuilder.password(env.getProperty("db.password"));
-        return dataSourceBuilder.build();
+        DriverManagerDataSource ds = new DriverManagerDataSource();
```
paveljandejsek: May I ask why the replacement? I thought the datasource builder still exists? Or is it not usable for us anymore due to some reason?

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
```diff
@@ -443,7 +443,7 @@ private void verifyResponseHeaders(HttpHeaders actualHeaders, ExamplesTable data
             if (key.startsWith(HEADER_START) && !key.equals(STATUS_HEADER)) {
                 String headerKey = key.substring(HEADER_START.length());
                 String assertionErrorMessage = "Headers don't contain " + headerKey + "\n" + actualResponseMessage;
-                assertThat(actualHeaders.containsKey(headerKey)).as(assertionErrorMessage).isTrue();
+                assertThat(actualHeaders.get(headerKey) != null).as(assertionErrorMessage).isTrue();
```
paveljandejsek: Nitpick, but wouldn't `containsHeader` be more readable in this case?
```suggestion
                assertThat(actualHeaders.containsHeader(headerKey)).as(assertionErrorMessage).isTrue();
```

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
```diff
@@ -177,7 +177,7 @@ private void storeResponse(final HttpStatusCodeException e) {
 
     private HttpEntity createRequestEntity(ExamplesTable data) throws IOException {
         if (data == null) { //return dummy
-            return new HttpEntity<>("", null);
+            return new HttpEntity<>("", HttpHeaders.EMPTY);
```
paveljandejsek, line 180: Just curious - did the old null cause issues? But yeah, this new version is definitely more readable 👍

jbehave-support-core/src/main/java/org/jbehavesupport/core/report/extension/RestXmlReporterExtension.java
```diff
@@ -159,10 +159,10 @@ private Map<String, String> getResponseMessageAttributes(RestMessageContext mess
         return responseMessageAttributes;
     }
 
-    private Map<String, String> getHeaderAttributes(Map.Entry<String, List<String>> header) {
+    private Map<String, String> getHeaderAttributes(String key, List<String> values) {
         Map<String, String> headerAttributes = new HashMap<>();
-        headerAttributes.put("key", header.getKey());
-        headerAttributes.put("value", header.getValue().get(0));
+        headerAttributes.put("key", key);
+        headerAttributes.put("value", values.isEmpty() ? "" : values.getFirst());
```
paveljandejsek: Again just curious - won't the empty string potentially cause any issues compared to the old behavior?
But otherwise nice job, it is more readable this way 🙂

jbehave-support-core/docs/Migration.md
```diff
@@ -16,14 +16,14 @@ We removed `org.jbehavesupport.core.internal.ConditionalOnMissingBean` annotatio
 ### Deprecated code removal
 We have removed all previously deprecated code - please refer to [Deprecated.md](Deprecated.md) for more details how to migrate.
 
-### Spring 6 upgrade
-We are currently using Spring 6 internally, this brings with it several other requirements:
+### Spring 7 upgrade
```
paveljandejsek, line 19: I would maybe keep the old one as it is and create a new section `Version 2.1.x` (maybe begs a question of framework upgrade does not warrant major 3.x version, but due to the limited breaking changes I would say 2.1.0? What do you think?)

README.md
```diff
@@ -18,7 +18,7 @@ along with support for [verification](jbehave-support-core/docs/General.md#verif
 [expression commands](jbehave-support-core/docs/Expression-commands.md) and 
 basic [reporting](jbehave-support-core/docs/Reporting.md).
 
-Currently supported Java versions are 17, 21 and 23 (latest LTS and latest version).
+Currently supported Java versions are 21 and 25 (latest LTS and latest version).
```
paveljandejsek: Can I ask what is the reason for dropping Java 17 support? As far as I know Spring Boot 4 and Spring 7 should still work with it? Or I am wrong? (haven't tried myself, just asking)

jbehave-support-core-test/jbehave-support-core-test-app/src/main/resources/application.yml
```diff
@@ -1,7 +1,8 @@
 spring:
+  jackson:
+    constructor-detector: explicit-only
```
paveljandejsek, line 3: May I ask why is explicit only need in our case?

jbehave-support-core-test/jbehave-support-core-test-app/src/main/resources/application.yml
```diff
@@ -10,6 +11,10 @@ spring:
       name: sa
       password: sa
 
+h2:
+  tcp:
+    port: 11112
```
paveljandejsek: Can I ask why was this moved from implicit spring datasource?

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/SecurityConfig.java
```diff
@@ -4,22 +4,27 @@
...
+import org.springframework.security.config.web.PathPatternRequestMatcherBuilderFactoryBean;
 import org.springframework.security.web.SecurityFilterChain;
 
-import static org.springframework.security.web.util.matcher.AntPathRequestMatcher.antMatcher;
-
 @Configuration
 public class SecurityConfig {
 
+    @Bean
+    PathPatternRequestMatcherBuilderFactoryBean requestMatcherBuilder() {
```
paveljandejsek: Can I ask why is this needed? I was under the impression the requestMatchers("xxx") work by default? Or was that changed somehow?

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/DatabaseConfig.java
```diff
@@ -31,4 +35,16 @@ public Server initServer() {
         }
         return h2Server;
     }
+
+    @Bean("TEST")
+    @DependsOn("initServer")
+    public DataSource dataSource() {
```
paveljandejsek: Do we really need this? Shouldn't auto configuration provide datasource for us from the spring.datasource properties automatically?

paveljandejsek (commented):
Nice job, thanks a lot!
I have mainly just questions and a couple of nitpicks 🙂 

Main points for me:

- What do we want the next version to be? 2.1.0 vs 3.0.0?
- What was the reason for Java 17 support being removed?
- The whole datasource stuff in the test app feels weird, I would hope autoconfig and liquibase to handle the heavy lifting instead of doing it manually 🙂 
- We should probably take a look at jbehave runner as well in some future PR (since Boot 4 brings in JUnit 6 which the runner currently does not support)

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/SecurityConfig.java
Mejdlo: After upgrade I was dealing with some errors since antMatchers has been deleted and after some googling, I have found this solution.

jbehave-support-core/src/main/java/org/jbehavesupport/core/report/extension/RestXmlReporterExtension.java
Mejdlo: I don't think that it should. It just some sort of a QA deformation with validation.

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
Mejdlo: The previous implementation is not usable anymore due to the latest changes and Intellj Idea has been screaming on me 🙂

jbehave-support-core/src/test/groovy/org/jbehavesupport/test/TestConfiguration.java
Mejdlo: If I am not mistaken there has been changes how it works in Spring Boot 4 and it causes issues H2 databases in tests.

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/DatabaseConfig.java
Mejdlo: You are right. This is not needed anymore. This is just my mess when I was trying to solve H2 tests. Sorry for that and it will be removed in the next commit.

jbehave-support-core-test/jbehave-support-core-test-app/src/main/resources/application.yml
Mejdlo: I decided to use this because of some issues with lombok after the update in some classes (@RequiredArgsConstructor, @AllArgsConstructor)

jbehave-support-core-test/jbehave-support-core-test-app/src/main/resources/application.yml
Mejdlo: It is not needed anymore. It was just some attempts during desperate times because of the H2 database tests. Will be removed in the next commit.

jbehave-support-core/docs/Migration.md
Mejdlo: I am used to ```major.minor.fix``` so I would prefer probably 3.0.0, since java and spring boot will be updated. But surely we could use 2.1.0 😄  Jakub has created version 3.0.0-RC for now to be able to test it properly in our test project (since we are probably the only team that is using it properly) but sure, we can use 2.1.0 after the release.

jbehave-support-core/src/main/java/org/jbehavesupport/core/rest/RestServiceHandler.java
Mejdlo: Sure, I can change that. Good point 👌

jbehave-support-core/src/test/resources/db/init.sql
Mejdlo: If I remember correctly, tests were failing for not knowing table Person even though it is mentioned in the changelog.

README.md
Mejdlo: Some of the changes were not compatible with Java 17 and since Java 21 is used for quite some time, I though it is not necessary to make new version of JBehave compatible with Java 17. But if you think we should still keep compatibility with Java 17, I can make it happen 🙂

Mejdlo:
> Nice job, thanks a lot! I have mainly just questions and a couple of nitpicks 🙂
> 
> Main points for me:
> 
> * What do we want the next version to be? 2.1.0 vs 3.0.0?
> * What was the reason for Java 17 support being removed?
> * The whole datasource stuff in the test app feels weird, I would hope autoconfig and liquibase to handle the heavy lifting instead of doing it manually 🙂
> * We should probably take a look at jbehave runner as well in some future PR (since Boot 4 brings in JUnit 6 which the runner currently does not support)

Thank you for your review ! 🙂 I tried my best and surely you can see that I am not on your java knowledge level. I hope that I have answered everything. As I wrote, I would prefer to use version 3.0.0 since there is update of spring boot and java. I definitely agree that there should come some other PR to update other things. And Compatibility with Java 21 ? I can make it happen, I just though that it won't be necessary and compatibility with Java 21 will be enough but it is your call. It would take just couple of small changes to make it compatible still with Java 17

pom.xml
Mejdlo: It is necessary to set atleast target due to gmavenplus, otherwise the tests won't start. But source can be deleted so it is removed in the next commit.

README.md
xsmrcek: I would prefer compability with Java 17 as it is still supported LTS verions of Java :)

pom.xml
paveljandejsek: And here I was blissfully forgetting about the joys of mixing groovy and java code in one codebase 😄 
Thanks for the explanation and investagation, could you please add a comment above it similar to the Lombok one on the compiler.proc above it? Just so in the future we know that its there because of gmavenplus

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/SecurityConfig.java
paveljandejsek: I had some issues with fully running the pipeline locally (my docker env is fucked atm), but when running the app as a standalone I was able to get it to run without this bean and with the remaining security config as is... Can you please try as well? Or let me know what the issue was/is so I could try it on my side as well?

jbehave-support-core/src/test/resources/db/init.sql
paveljandejsek: I think the issue might be with the change of starters, see:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide

image

If I add the `spring-boot-starter-liquibase` dependency (liquibase-core is no longer needed) I can see the migrations running, so the init script should not be needed (at least I hope so)🤞

paveljandejsek:
> > Nice job, thanks a lot! I have mainly just questions and a couple of nitpicks 🙂
> > Main points for me:
> > 
> > * What do we want the next version to be? 2.1.0 vs 3.0.0?
> > * What was the reason for Java 17 support being removed?
> > * The whole datasource stuff in the test app feels weird, I would hope autoconfig and liquibase to handle the heavy lifting instead of doing it manually 🙂
> > * We should probably take a look at jbehave runner as well in some future PR (since Boot 4 brings in JUnit 6 which the runner currently does not support)
> 
> Thank you for your review ! 🙂 I tried my best and surely you can see that I am not on your java knowledge level. I hope that I have answered everything. As I wrote, I would prefer to use version 3.0.0 since there is update of spring boot and java. I definitely agree that there should come some other PR to update other things. And Compatibility with Java 21 ? I can make it happen, I just though that it won't be necessary and compatibility with Java 21 will be enough but it is your call. It would take just couple of small changes to make it compatible still with Java 17

No need to be modest, you probably code more than I do lately, as you can see half of the stuff from me were questions how the stuff even works nowadays 😄 
And sorry for slow turnaround from my side, been a bit busier than I would like lately...
Anyway if we are in the end able to keep Java 17 around then I would probably do 2.1.0 since we should not be introducing too many breaking changes. if not then 3.0.0 sounds like a plan.

I have put a couple more replies to the comments, I still want to go over the H2 stuff a bit more on my side (and fix my env to be able to run clean verify without errors locally, but that's my issue)  - hopefully tomorrow evening I should have the rest of the replies for you

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/SecurityConfig.java
paveljandejsek: Ok so the docker being borked was actually caused by testcontainers - do you think you could remove the `version.testcontainers` and the testcontainers dependency (line 202) from the main pom? Boot 4 now seems to include testcontainers bom (and a much newer version than we had previously), so we can rely on the version from the spring boot now.... (Or if you want I can do it in my own commit either to your branch/this PR, or I can do it after we merge this in a separate PR, up to you)

jbehave-support-core/src/test/groovy/org/jbehavesupport/test/TestConfiguration.java
paveljandejsek: Ok so in the end I was able to make it work with the previous code, the issue was the same as with the liquibase - i.e. instead of `spring-jdbc` dependency in jbehave-support-core pom we now need the starter one:
```
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-jdbc</artifactId>
        </dependency>
```
image

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide

paveljandejsek:
Sorry for being late, the docker issue for me took longer than it should have (OS upgrade, docker updates, fiddling around with some random stuff before I thought about checking testcontainers themselves 🤦‍♂️), I put the remaining comments in there... Thank you again for the nice work!

Mejdlo:
> Sorry for being late, the docker issue for me took longer than it should have (OS upgrade, docker updates, fiddling around with some random stuff before I thought about checking testcontainers themselves 🤦‍♂️), I put the remaining comments in there... Thank you again for the nice work!

Thank you for the latest comments. I hope that I have changed everything, that was mentioned 😄  I have kept java 17 in the project and made it compatible so I think, that version 2.1.0 would be valid

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           name: reports-java-${{ matrix.java }}
           path: jbehave-support-core/target/reports
       - name: Upload test coverage for integration tests
-        if: matrix.java == 17
+        if: matrix.java == 25
```
paveljandejsek:
```suggestion
        if: matrix.java == 17
```
We need same version for sonar as we are uploading the artifacts for (just a rollback of probably forgotten changes)

.github/workflows/ci.yml
```diff
@@ -115,10 +115,10 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-
-      - name: Set up Java 17
+      - name: Set up Java 21
```
paveljandejsek:
```suggestion
      - name: Set up Java 17
```
We need same version for sonar as we are uploading the artifacts for (just a rollback of probably forgotten changes)

.github/workflows/ci.yml
```diff
@@ -115,10 +115,10 @@ jobs:
...
         uses: actions/setup-java@v4
         with:
-          java-version: 17
+          java-version: 21
```
paveljandejsek:
```suggestion
          java-version: 17
```
We need same version for sonar as we are uploading the artifacts for (just a rollback of probably forgotten changes)

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/DatabaseConfig.java
```diff
@@ -20,8 +21,7 @@ public class DatabaseConfig {
     public Server initServer() {
         Server h2Server;
         try {
-            h2Server = Server.createTcpServer("-ifNotExists", "-tcpAllowOthers", "-tcpPort", env.getProperty("spring.datasource.port")).start();
-            if (h2Server.isRunning(true)) {
+            h2Server = Server.createTcpServer("-ifNotExists", "-tcpAllowOthers", "-tcpPort", env.getProperty("spring.datasource.port")).start();            if (h2Server.isRunning(true)) {
```
paveljandejsek: Nitpick: I think there is a mess up formatting in here, probably a mistake?

jbehave-support-core-test/jbehave-support-core-test-app/pom.xml
```diff
@@ -118,7 +118,7 @@
                             <image>
                                 <tags>docker.io/library/jbehave-support-core-test-app:latest</tags>
                                 <env>
-                                    <BP_JVM_VERSION>17</BP_JVM_VERSION>
+                                    <BP_JVM_VERSION>21</BP_JVM_VERSION>
```
paveljandejsek:
```suggestion
                                    <BP_JVM_VERSION>17</BP_JVM_VERSION>
```
We need the lowest version here for pipeline on 17 to work I think? (basically just a rollback of probably forgotten changes?)

jbehave-support-core/docs/Migration.md
```diff
@@ -16,14 +16,14 @@ We removed `org.jbehavesupport.core.internal.ConditionalOnMissingBean` annotatio
...
+We are currently using Spring 7 internally, this brings with it several other requirements:
 
-> #### Java 17
->We now currently require Java 17 as a minimal Java version (previously 8).
+> #### Java 25
```
paveljandejsek:
```suggestion
> #### Java 17
```

jbehave-support-core/src/test/groovy/org/jbehavesupport/test/TestConfiguration.java
```diff
@@ -20,14 +20,18 @@
 import org.jbehavesupport.test.support.TestWebServiceHandler;
 import org.openqa.selenium.MutableCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.SpringBootConfiguration;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.jdbc.DataSourceBuilder;
+import org.springframework.jdbc.datasource.DriverManagerDataSource;
```
paveljandejsek: I think a couple of forgotten unused imports?

jbehave-support-core/src/test/resources/test.yml
```diff
@@ -10,7 +10,7 @@ rest:
   password: sa
 
 db:
-  url: jdbc:h2:tcp://localhost:11112/mem:test;MODE=ORACLE
+  url: jdbc:h2:mem:test;MODE=ORACLE;DB_CLOSE_DELAY=-1
```
paveljandejsek:
```suggestion
  url: jdbc:h2:tcp://localhost:11112/mem:test;MODE=ORACLE
```
I think we now need the previous version back? Since the `TestAppContainer` uses fixed port...

paveljandejsek (changes requested):
Apologies, but I found a couple more nitpicks (tried to put suggestions where I could), mostly probably just forgotten stuff from the editing I think... But otherwise I think we are good to go, once you manage to solve them I will try to get the new version 2.1.0 out once we have it merged 🙂 

Please don't hate me 😄

jbehave-support-core-test/jbehave-support-core-test-app/src/main/java/org/jbehavesupport/core/test/app/DatabaseConfig.java
Mejdlo: You are right. Somehow I missed that while I was doing latest changes. Sorry for that 😞

Mejdlo:
> Apologies, but I found a couple more nitpicks (tried to put suggestions where I could), mostly probably just forgotten stuff from the editing I think... But otherwise I think we are good to go, once you manage to solve them I will try to get the new version 2.1.0 out once we have it merged 🙂
> 
> Please don't hate me 😄

No need for hate. Those were my mistakes that I have forgot to return back settings while I was trying to run those tests locally. 😄  Sorry for that. I think that everything should be (hopefully) correct for now. And somehow optimalization of unused imports was not marked (even though I was using it. Weird 😄 )

paveljandejsek (approved):

paveljandejsek:
Looking good, thank you very much! Approved from my side, I will try to get it merged and released once I am able.

xsmrcek (approved):

